PR #107, opened 2020-01-08
fix: Return 404 when feature or experiment not found in some routes

mjc1283:
Update the following routes to return a 404 response with no content when the feature or experiment with the key given in the URL path doesn't exist:

```
GET /features/featureKey
GET /experiments/experimentKey
GET /users/userId/features/featureKey
GET /users/userId/experiments/experimentKey
POST /users/userId/features/featureKey
POST /users/userId/experiments/experimentKey
```

This is implemented with two new methods of `CachedOptlyMiddleware` in the `middleware` package: `FeatureCtx` and `ExperimentCtx`.

In addition to returning 404 early, when the entity is found, it's set in request context for handlers later in the chain. The final handlers for the above routes are updated to access the entity through request context, instead of directly accessing URL params.

Also changed:
- Change response status from 422 to 500 in all handlers that rely on middleware to populate `OptlyClient` or `OptlyContext`, for cases when the context failed to be set properly. This is for consistency with the experiment and feature handlers updated, and the `FeatureCtx `and `ExperimentCtx` middlewares.


pkg/handlers/user.go
```diff
@@ -110,8 +119,14 @@ func (h *UserHandler) GetVariation(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	experimentKey := chi.URLParam(r, "experimentKey")
-	renderVariation(w, r, experimentKey, false, optlyClient, optlyContext)
+	experiment, err := middleware.GetExperiment(r)
+	if err != nil {
+		middleware.GetLogger(r).Error().Err(err).Msg("Calling middleware GetExperiment")
+		RenderError(err, http.StatusInternalServerError, w, r)
```
mikecdavis: We have been returning a `http.StatusUnprocessableEntity` (422) if the context wasn't properly populated. (I think I stole that response from a middleware example) Here we're returning a 500. 500 makes sense since if we got here it's probably due to a mis configuration on the server side. Can you make the others return a 500 as well for consistency?

pkg/middleware/cached.go
```diff
@@ -103,3 +110,74 @@ func (ctx *CachedOptlyMiddleware) UserCtx(next http.Handler) http.Handler {
...
+		case err == nil:
+			GetLogger(r).Debug().Str("experimentKey", experimentKey).Msg("Added experiment to request context in ExperimentCtx")
+			ctx := context.WithValue(r.Context(), OptlyExperimentKey, &experiment)
+			next.ServeHTTP(w, r.WithContext(ctx))
+			return
+		case errors.Is(err, optimizely.ErrEntityNotFound):
+			statusCode = http.StatusNotFound
+		default:
+			statusCode = http.StatusInternalServerError
+		}
```
mikecdavis: Nice use of error handling. I'm not sure if this is the first time, but this does limit Agent to Go version 1.13. Something to consider.

mjc1283: The only drawback I can think of is if we want to allow consuming Agent as a library, then the users need to be on 1.13.

PR #121, opened 2020-01-13
feat: using metrics registry to get all metrics

pawels-optimizely:


pkg/metrics/metrics.go
```diff
@@ -86,6 +91,42 @@ func (m *Registry) GetGauge(key string) metrics.Gauge {
...
+	combinedKey := TimerPrefix + "." + key
+
+	m.counterLock.Lock()
+	defer m.counterLock.Unlock()
+	if val, ok := m.metricsCounterVars[combinedKey]; ok {
+		return val
+	}
+
+	return m.createCounter(combinedKey)
+}
```
mikecdavis: We should define `Timer` on the `MetricsRegistry` as an abstract construct and not just an alias for a `Counter`. This moves this logic out of this middleware implementation which has the benefit of being reusable directly out of the registry and removes the go-kit dependency from that module.
```go
type Timer struct {
	hits  go_kit_metrics.Counter
	total go_kit_metrics.Counter
	hist  go_kit_metrics.Histogram
}

func (m *Registry) NewTimer(key string) *Timer {
	...
	name := TimerPrefix + "." + key
	return &Timer {
		hits:  m.createCounter(name + ".hits"),
		total: m.createCounter(name + ".total"),
		hist:  m.createHistogram(name+".hist"),
	}
}

func (t *Timer) Update(delta float64) {
	t.hitCounts.Add(1)
	t.responseTime.Add(timeDiff)
	t.responseTimeHist()
}
```

pkg/middleware/metrics.go
```diff
@@ -22,39 +22,35 @@ import (
...
+		ResponseTimeHistogram: metricsRegistry.GetHistogram(metrics.TimerPrefix, key+".responseTimeHist"),
 	}
 }
 
 // Metricize updates counts, total response time, and response time histogram
 // for each URL hit, key being a combination of a method and route pattern
-func Metricize(key string) func(http.Handler) http.Handler {
-	singleMetric := NewMetrics(key)
+func Metricize(key string, metricsRegistry *metrics.Registry) func(http.Handler) http.Handler {
+	singleMetric := NewMetrics(key, metricsRegistry)
```
mikecdavis: With the proposed change to the registry, this becomes:
```go
timer := metricsRegistry.NewTimer(key)
```

pkg/metrics/metrics.go
pawels-optimizely: I have made this change, I like my previous implementation - I don't need to make a map specially for timers. Take a look...

PR #124, opened 2020-01-14
Add Agent version and name.

mikecdavis:
## Summary
* Set's the ClientName and Version as part of the build

Client name and version are included in the log requests to Optimizely to track adoption. We want to be able to distinguish Agent requests from native go-sdk requests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,8 +23,7 @@ MAKEFLAGS += --silent
 # Use linker flags to strip debugging info from the binary.
 # -s Omit the symbol table and debug information.
 # -w Omit the DWARF symbol table.
-LDFLAGS=-ldflags "-s -w -X main.Version=${APP_VERSION}"
-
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -23,8 +23,7 @@ MAKEFLAGS += --silent
 # Use linker flags to strip debugging info from the binary.
 # -s Omit the symbol table and debug information.
 # -w Omit the DWARF symbol table.
-LDFLAGS=-ldflags "-s -w -X main.Version=${APP_VERSION}"
-
+LDFLAGS=-ldflags "-s -w -X main.Version=${APP_VERSION} -X github.com/optimizely/go-sdk/pkg/event.ClientName=Agent -X github.com/optimizely/go-sdk/pkg/event.Version=${APP_VERSION}"
```
pawels-optimizely: why do we need to set  pkg/event variables?

PR #128, opened 2020-01-20
feat (notification): initial copy of the event stream for decision notifications

thomaszurkan-optimizely:
##Summary
This is the intial implementation of the event stream.  I don't think metrics and timers should be part of this endpoint but I did this to get it going.  Endpoints are as follow:
1. Run make run
2. curl -i -X POST -H "X-Optimizely-SDK-Key: FeEqVPw2ZKLcxjHX5A732L" http://localhost:8080/users/tom/experiments/background_experiment
3. curl -N --http2 -H "Accept:text/event-stream" -H "X-Optimizely-SDK-Key: FeEqVPw2ZKLcxjHX5A732L" http://localhost:8080/notifications/event-stream

Output looks like the following:

data: {"Type":"ab-test","UserContext":{"ID":"tom","Attributes":{}},"DecisionInfo":{"experimentKey":"background_experiment","variationKey":"variation_b"}}


pkg/handlers/event_stream.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// Closed client connections
+	closingClients chan MessageChan
+
+	// Client connections registry
+	clients map[MessageChan]bool
+}
+
+// Listen on different channels and act accordingly
+func (broker *EventStreamBroker) listen() {
```
mikecdavis: Per our conversation, the "broker" here is not necessary since we've elected to add new listeners with each connection. This approach puts the burden on the go-sdk to scalably handle listeners being added and removed and reliably subscribed to. Addressing the heavy lifting from that side will ensure a robust implementation on the go-sdk side. Effectively the NotificationCenter replaces the need for this Broker.

pkg/handlers/event_stream.go
```diff
@@ -0,0 +1,161 @@
...
+	rw.Header().Set("Connection", "keep-alive")
+	rw.Header().Set("Access-Control-Allow-Origin", "*")
+
+	// Each connection registers its own message channel with the EventStreamBroker's connections registry
+	messageChan := make(MessageChan)
+
+	// Signal the broker that we have a new connection
+	broker.newClients <- messageChan
+
+	id,err := optlyClient.DecisionService.OnDecision(func(decision notification.DecisionNotification) {
```
mikecdavis: We should be able to use the `registry` to fetch NotificationCenter instances by SDK key.
```go
nc := registry.GetNotificationCenter(sdkKey)
nc.AddHandler(...)
```

pkg/handlers/event_stream.go
```diff
@@ -0,0 +1,109 @@
...
+		// The flush will fail if the connection is closed.  That will cause the handler to exit.
+		flusher.Flush()
+
+		log.Println("flushing")
+	}
+
+}
+
+// EventStreamHandler factory
+func NewEventStreamHandler() (esh *EventStreamHandler) {
```
mikecdavis: Is this constructor needed?

pkg/handlers/event_stream.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// Remove the decision listener if we exited.
+	defer func() {
+		optlyClient.DecisionService.RemoveOnDecision(id)
+	}()
+
+	// "raw" query string option
+	// If provided, send raw JSON lines instead of SSE-compliant strings.
+	req.ParseForm()
+	raw := len(req.Form["raw"]) > 0
```
mikecdavis: Should this be:
```
raw := req.URL.Query().Get("raw")
```

pkg/handlers/event_stream.go
```diff
@@ -0,0 +1,109 @@
...
+		<-notify
+		optlyClient.DecisionService.RemoveOnDecision(id)
+	}()
+
+	// block waiting or messages broadcast on this connection's messageChan
+	for {
+		// Write to the ResponseWriter
+		if raw {
+			// Raw JSON events, one per line
+			fmt.Fprintf(rw, "%s\n", <-messageChan)
```
mikecdavis: I think we're missing coverage here.

pkg/handlers/event_stream.go
thomaszurkan-optimizely: no, the parse form parses both the url query or the form based on whether it is a get or post.

.gitignore
```diff
@@ -15,3 +15,4 @@ generated/
 
 # Dependency directories (remove the comment below to include it)
 # vendor/
+.idea
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: that didn't work for me.  do i have to do something else as well?

pkg/handlers/event_stream.go
thomaszurkan-optimizely: I'd like to use a constructor so that if implementation changes.  It would all be in the background.

pkg/routers/api.go
```diff
@@ -91,22 +98,28 @@ func NewAPIRouter(opt *APIOptions) *chi.Mux {
...
 
-	r.Use(middleware.SetTime)
-	r.Use(render.SetContentType(render.ContentTypeJSON), middleware.SetRequestID)
+	r.Route("/notifications/event-stream", func(r chi.Router) {
+		r.Use(opt.middleware.ClientCtx)
+		r.Get("/", opt.eventsAPI.HandleEventSteam)
+	})
 
 	r.Route("/features", func(r chi.Router) {
+		setMiddleWareTime(r)
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: Nope.  All middleware must be added before routes.

pkg/handlers/event_stream.go
thomaszurkan-optimizely: added a unit test

pkg/handlers/event_stream_test.go
```diff
@@ -0,0 +1,212 @@
...
+	expected := "data: {\"Type\":\"feature\",\"UserContext\":{\"ID\":\"testUser\",\"Attributes\":{}},\"DecisionInfo\":{\"feature\":{\"featureEnabled\":true,\"featureKey\":\"one\",\"source\":\"feature-test\",\"sourceInfo\":{\"experimentKey\":\"1\",\"variationKey\":\"2\"}}}}\n\n"
+
+	suite.Equal(http.StatusOK, rec.Code)
+
+	// Unmarshal response
+	response := string(rec.Body.Bytes())
+	suite.Equal(expected,response)
+}
+
+func (suite *EventStreamTestSuite) TestActivateExperiment() {
```
mikecdavis: I think if we invert the goroutine and use a `sync.WaitGroup` then the tests will be a bit easier to manage and we don't need to rely on `sleep` statements which are inherently flaky. 
For example:
```go
func (suite *EventStreamTestSuite) TestActivateExperiment() {
	testVariation := suite.tc.ProjectConfig.CreateVariation("variation_a")
	suite.tc.AddExperiment("one", []entities.Variation{testVariation})

	req := httptest.NewRequest("GET", "/notifications/event-stream", nil)
	rec := httptest.NewRecorder()
	// create a cancelable request context
	ctx, cancelFun := context.WithCancel(req.Context())
	expected := `data: {"Type":"ab-test","UserContext":{"ID":"testUser","Attributes":{}},"DecisionInfo":{"experimentKey":"one","variationKey":"variation_a"}}` + "\n\n"

	wg := sync.WaitGroup{}
	wg.Add(1)

	go func() {
		suite.mux.ServeHTTP(rec, req.WithContext(ctx))
		suite.Equal(http.StatusOK, rec.Code)

		// Unmarshal response
		response := string(rec.Body.Bytes())
		suite.Equal(expected, response)
		wg.Done()
	}()

	suite.tc.OptimizelyClient.Activate("one", entities.UserContext{"testUser", make(map[string]interface{})})
	cancelFun()

	wg.Wait()
}
```

Similar comment on all the tests relying on `time.Sleep` in this suite.

.gitignore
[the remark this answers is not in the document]
mikecdavis: hm. Let's remove for now this and address independent of this PR.

pkg/handlers/interface.go
```diff
@@ -33,6 +33,11 @@ type ExperimentAPI interface {
 	ListExperiments(w http.ResponseWriter, r *http.Request)
 }
 
+// NotificationsAPI defines support for notification via server side events
+type NotificationsAPI interface {
```
mikecdavis: super nit, but other interfaces in this package use the singular tense. (e.g. `NotificationAPI`)

pkg/handlers/event_stream_test.go
mikecdavis: this might require another WaitGroup to be super certain the request was initiated, but the concept is sound.

pkg/handlers/event_stream.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		RenderError(err, http.StatusUnprocessableEntity, rw, req)
+		return
+	}
+
+	if optlyClient == nil {
+		e := fmt.Errorf("optlyContext not available")
+		RenderError(e, http.StatusUnprocessableEntity, rw, req)
+		return
+	}
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: It crashes my unit test without it.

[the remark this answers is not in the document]
mikecdavis: I see, your test is explicitly setting a `nil` value in the request context which _shouldn't_ happen, but it's a good check to be defensive. Can you move this check to the `GetOptlyClient` util method? Returning a non-nil `err` if the client is nil.

pkg/handlers/notification.go
```diff
@@ -0,0 +1,136 @@
...
+		e := fmt.Errorf("optlyContext not available")
+		RenderError(e, http.StatusUnprocessableEntity, w, r)
+		return
+	}
+
+	// Set the headers related to event streaming.
+	w.Header().Set("Content-Type", "text/event-stream")
+	w.Header().Set("Cache-Control", "no-cache")
+	w.Header().Set("Connection", "keep-alive")
+	w.Header().Set("Access-Control-Allow-Origin", "*")
```
mjc1283: Should `Access-Control-Allow-Origin` header value be configurable?

pkg/handlers/notification.go
```diff
@@ -0,0 +1,136 @@
...
+		err := optlyClient.DecisionService.RemoveOnDecision(id)
+		if err != nil {
+			middleware.GetLogger(r).Error().AnErr("removingOnDecision", err)
+		}
+	}()
+
+	// "raw" query string option
+	// If provided, send raw JSON lines instead of SSE-compliant strings.
+	_ = r.ParseForm()
+	raw := len(r.Form["raw"]) > 0
```
mjc1283: Is this a better way to check presence of `raw` in the query string (maybe more direct/readable)?
```go
raw := r.URL.Query().Get("raw") != ""
```

pkg/handlers/notification.go
```diff
@@ -0,0 +1,136 @@
...
+			// The flush will fail if the connection is closed.  That will cause the handler to exit.
+			flusher.Flush()
+			fmt.Println("received message", msg)
+		// Remove the decision listener if the connection is closed and exit
+		case sig := <-notify:
+			err := optlyClient.DecisionService.RemoveOnDecision(id)
+			if err != nil {
+				middleware.GetLogger(r).Error().AnErr("removingOnDecision", err)
+			}
+			fmt.Println("received close on the request.  So, we are shutting down this handler", sig)
```
mjc1283: Should use logger instead of `fmt.Println`? Or remove?

pkg/handlers/notification.go
```diff
@@ -0,0 +1,168 @@
...
+			}
+		}
+	}
+
+	return notificationsToAdd
+}
+
+// HandleEventSteam implements the http.Handler interface.
+// This allows us to wrap HTTP handlers (see auth_handler.go)
+// http://golang.org/pkg/net/http/#Handler
```
mikecdavis: Please remove

PR #129, opened 2020-01-23
Refactor overrides into it's own API set

mikecdavis:
## Summary
* Moves user overrides into into own file and test file.
* Updates the method signture to use a request body as opposed to a URL param

After some discussion with @mjc1283 we decided to refactor the overrides API. This makes the API explicitly disjoint from the typical API usage by namespacing it under `/overrides`. We also changed the method signature to better reflect that only a single override can exist for a given experiment/user combo.

### DIFF
--- a/pkg/handlers/user_overrides_test.go
+++ b/pkg/handlers/user_overrides_test.go
@@ -1,0 +1,169 @@
+/****************************************************************************
+ * Copyright 2020, Optimizely, Inc. and contributors                        *
+ *                                                                          *
+ * Licensed under the Apache License, Version 2.0 (the "License");          *
+ * you may not use this file except in compliance with the License.         *
+ * You may obtain a copy of the License at                                  *
+ *                                                                          *
+ *    http://www.apache.org/licenses/LICENSE-2.0                            *
+ *                                                                          *
+ * Unless required by applicable law or agreed to in writing, software      *
+ * distributed under the License is distributed on an "AS IS" BASIS,        *
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *
+ * See the License for the specific language governing permissions and      *
+ * limitations under the License.                                           *
+ ***************************************************************************/
+
+// Package handlers //
+package handlers
+
+import (
+	"net/http"
+	"net/http/httptest"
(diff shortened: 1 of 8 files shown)


pkg/routers/api.go
```diff
@@ -104,8 +106,13 @@ func NewAPIRouter(opt *APIOptions) *chi.Mux {
...
 		r.With(getVariationTimer, opt.middleware.ExperimentCtx).Get("/experiments/{experimentKey}", opt.userAPI.GetVariation)
 		r.With(activateExperimentTimer, opt.middleware.ExperimentCtx).Post("/experiments/{experimentKey}", opt.userAPI.ActivateExperiment)
-		r.With(setForcedVariationTimer).Put("/experiments/{experimentKey}/variations/{variationKey}", opt.userAPI.SetForcedVariation)
-		r.With(removeForcedVariationTimer).Delete("/experiments/{experimentKey}/variations", opt.userAPI.RemoveForcedVariation)
+	})
+
+	r.Route("/overrides/users/{userID}", func(r chi.Router) {
+		r.Use(opt.middleware.ClientCtx, opt.middleware.UserCtx)
+
+		r.With(setForcedVariationTimer).Put("/experiments/{experimentKey}/variations/{variationKey}", opt.userOverrideAPI.SetForcedVariation)
```
pawels-optimizely: according to our conversation over slack, you said variationKey should be in the PUT body. Also, do we need /variations/ in the URL ?

mikecdavis: well that's embarrassing, I updated everything but the api router 🤦‍♂

pkg/handlers/user_overrides_test.go
```diff
@@ -0,0 +1,178 @@
...
+	"net/http/httptest"
+	"testing"
+
+	"github.com/optimizely/go-sdk/pkg/decision"
+
+	"github.com/optimizely/agent/pkg/optimizely"
+	"github.com/optimizely/agent/pkg/optimizely/optimizelytest"
+
+	"github.com/go-chi/chi"
+	"github.com/optimizely/go-sdk/pkg/entities"
```
pawels-optimizely: nit: group go-sdk/pkg imports together

PR #130, opened 2020-01-23
Add type coercion to user attributes. 

mikecdavis:
## Summary
* Coerce user attributes to support data types
* Update tests to match coerced values

This change fixes a compatibility gap where all user attribute values were being handled as strings. Values are now coerced into typed values when applicable. Note that this will require some documentation on how values are coerced since this is unique to Agent.

Note that strings can be forced if quoted in the URL params.

### DIFF
--- a/pkg/middleware/cached.go
+++ b/pkg/middleware/cached.go
@@ -87,7 +87,6 @@ func (mw *CachedOptlyMiddleware) ClientCtx(next http.Handler) http.Handler {
 // detail from a UPS or attribute store.
 func (mw *CachedOptlyMiddleware) UserCtx(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-
 		userID := chi.URLParam(r, "userID")
 		if userID == "" {
 			RenderError(fmt.Errorf("invalid request, missing userId"), http.StatusBadRequest, w, r)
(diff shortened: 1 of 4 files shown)


pkg/middleware/cached.go
```diff
@@ -87,26 +87,28 @@ func (mw *CachedOptlyMiddleware) ClientCtx(next http.Handler) http.Handler {
...
+		logger := GetLogger(r)
 		values := r.URL.Query()
 		attributes := make(map[string]interface{})
+
 		for k, v := range values {
 			// Assuming a single KV pair exists in the query parameters
-			attributes[k] = v[0]
-			GetLogger(r).Debug().Str("attrKey", k).Str("attrVal", v[0]).Msg("User attribute.")
+			val := v[0]
+			attributes[k] = CoerceType(val)
```
pawels-optimizely: I thought we will be providing an option to do that, not all the time.

mikecdavis: This is for attributes in the request we should always coerce these. The optional typing is on the response for feature variables which will be part of a different PR.

pkg/middleware/utils.go
```diff
@@ -86,3 +87,27 @@ func GetExperiment(r *http.Request) (*config.OptimizelyExperiment, error) {
...
+
+	if i, err := strconv.ParseInt(s, 10, 0); err == nil {
+		return i
+	}
+
+	if d, err := strconv.ParseFloat(s, 64); err == nil {
+		return d
+	}
+
+	if b, err := strconv.ParseBool(s); err == nil {
```
pawels-optimizely: nit: knowing that ParseBool accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False, I think maybe we should limit inputs to true/false only , otherwise looks good

PR #132, opened 2020-01-24
Refactor configuration struct and add /config endpoint.

mikecdavis:
## Summary
* Move Author, Name and Version to root config
* Move SDKKeys and Processor to root config
* Add uptime to /info because it seamed like a good idea
* Add admin /config route to support operations and validate config
* Updates README.md and config.yaml

Exposing "optly" in the configs was gross. Moving items to the root of the application config felt more appropriate in some cases as Agent _is_ optimizely. I also wanted to expose the configuration in the routes as this is a handy endpoint when you want to confirm that certain configs are loaded etc.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -104,7 +104,8 @@ func main() {
 
 	ctx, cancel := context.WithCancel(context.Background()) // Create default service context
 	sg := server.NewGroup(ctx, conf.Server)                 // Create a new server group to manage the individual http listeners
-	optlyCache := optimizely.NewCache(ctx, conf.SDKKeys, sdkMetricsRegistry)
+	optlyCache := optimizely.NewCache(ctx, conf.Processor, sdkMetricsRegistry)
+	optlyCache.Init(conf.SDKKeys)
 
 	// goroutine to check for signals to gracefully shutdown listeners
 	go func() {
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -104,7 +104,7 @@ func main() {
 
 	ctx, cancel := context.WithCancel(context.Background()) // Create default service context
 	sg := server.NewGroup(ctx, conf.Server)                 // Create a new server group to manage the individual http listeners
-	optlyCache := optimizely.NewCache(ctx, conf.Optly, sdkMetricsRegistry)
+	optlyCache := optimizely.NewCache(ctx, conf.SDKKeys, sdkMetricsRegistry)
 
 	// goroutine to check for signals to gracefully shutdown listeners
 	go func() {
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -46,7 +46,7 @@ func initConfig(v *viper.Viper) error {
 	// Load defaults from the AgentConfig by loading the marshaled values as yaml
 	// https://github.com/spf13/viper/issues/188
 	defaultConf := config.NewDefaultConfig()
-	defaultConf.Admin.Version = Version
+	defaultConf.Version = Version
 	b, err := yaml.Marshal(defaultConf)
 	if err != nil {
 		return err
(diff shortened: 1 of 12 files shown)


cmd/main.go
```diff
@@ -117,10 +118,10 @@ func main() {
 		cancel()
 	}()
 
-	log.Info().Str("version", conf.Admin.Version).Msg("Starting services.")
+	log.Info().Str("version", conf.Version).Msg("Starting services.")
 	sg.GoListenAndServe("api", conf.API.Port, routers.NewDefaultAPIRouter(optlyCache, conf.API, agentMetricsRegistry))
 	sg.GoListenAndServe("webhook", conf.Webhook.Port, routers.NewWebhookRouter(optlyCache, conf.Webhook))
-	sg.GoListenAndServe("admin", conf.Admin.Port, routers.NewAdminRouter(conf.Admin)) // Admin should be added last.
+	sg.GoListenAndServe("admin", conf.Admin.Port, routers.NewAdminRouter(*conf)) // Admin should be added last.
```
mjc1283: Should `NewAdminRouter` accept a pointer type, instead of dereferencing it here?

config/config.go
```diff
@@ -60,62 +59,64 @@ func NewDefaultConfig() *AgentConfig {
...
+
+	Admin     AdminConfig     `yaml:"admin" json:"admin"`
+	API       APIConfig       `yaml:"api" json:"api"`
+	Log       LogConfig       `yaml:"log" json:"log"`
+	Processor ProcessorConfig `yaml:"processor" json:"processor"`
+	Server    ServerConfig    `yaml:"server" json:"server"`
+	Webhook   WebhookConfig   `yaml:"webhook" json:"webhook"`
 }
 
 // OptlyConfig holds the set of SDK keys to bootstrap during initialization
```
mjc1283: Need to update this comment

config/config.go
```diff
@@ -60,62 +59,64 @@ func NewDefaultConfig() *AgentConfig {
...
-	SDKKeys   []string        `yaml:"sdkkeys"`
+	Processor ProcessorConfig `yaml:"processor" json:"processor"`
 }
 
 // ProcessorConfig holds the configuration options for the Optimizely Event Processor.
 type ProcessorConfig struct {
-	BatchSize     int           `yaml:"batchSize" default:"10"`
-	QueueSize     int           `yaml:"queueSize" default:"1000"`
-	FlushInterval time.Duration `yaml:"flushInterval" default:"30s"`
+	BatchSize     int           `yaml:"batchSize" json:"batchSize" default:"10"`
```
mjc1283: Nit: any lint or test we can set up to ensure both `yaml` and `json` remappings on everything?

pkg/optimizely/cache.go
```diff
@@ -38,12 +38,12 @@ type OptlyCache struct {
 	optlyMap        cmap.ConcurrentMap
 	ctx             context.Context
 	wg              sync.WaitGroup
-	conf            config.OptlyConfig
+	conf            config.ProcessorConfig
```
mjc1283: Should we rename this field to `processorConf`?

config/config.go
mikecdavis: I can see if these two can be consolidated.

pkg/optimizely/cache.go
mikecdavis: no objections

config/config.go
mikecdavis: I couldn't find one, but since `yaml` is only used for unmarshaling then we don't technically need to map the field names. We do however need the json field mappings to yield the desired casing in the `/config` endpoint so I'll remove the yaml to avoid any potential discrepancies.

cmd/main.go
mikecdavis: I was going back and forth on this. I wanted to be consistent with all of the the other constructors which are not using a pointer.

PR #133, opened 2020-01-25
New /activate API proposal.

mikecdavis:
## Summary
* Consolidate experiment and feature "decision" APIs into a single `/activate` endpoint
* Move event tracking endpoint to `/track/{eventKey}` (previously `.../events/{eventKey}`)
* Move all "user" data points into the request body
* Query parameters will be reversed for selection and filtering criteria.

Including potentially sensitive information in the URL is a privacy concern. Web servers, proxies and gateways log URLs so we should not include potentially compromising data in the URL if it can be avoided.

## Trade-offs
Q: Why `/activate`?
A: While "activate" has not been the most loved name for our SDK APIs, it's an entrenched verb in the full stack lexicon. So instead of attempting to define new terminology we will consolidate on "activate" for Experiments _and_ Features. Any adjustments to SDK terminology will be made in conjunction with the SDK APIs if we decide to pursue a change in the future.

Q: Why consolidate experiment and feature decisions?
A: We are trying to optimize cleanliness in the API. Given that the action is basically the same whether we're deciding on an experiment variation or feature availability or a variation of a feature, the distinction is not required at the "action" level. Also the API responses are so similar that consolidating them makes implementing (or generating) API clients much simpler.


mikecdavis:
After speaking with @ceimaj some more on this. We might want to use `/activate` to avoid introducing yet another verb into the documentation, but instead consolidating the "activate" action to apply to both features and experiments. This might provide a better migration path as opposed to making things worse.

pkg/handlers/activate.go
```diff
@@ -0,0 +1,142 @@
+/****************************************************************************
```
asaschachar: The description of this PR is all about adding a /decide API, isn't this adding a /activate api instead?

asaschachar: nvm, just saw your comment in the PR. Maybe update the main PR description to minimize confusion?

docs/getting-started.md
```diff
@@ -38,35 +38,26 @@ s.headers.update({'X-Optimizely-SDK-Key': '<<YOUR-SDK-KEY>>'})
...
-
-print(resp.json())
+for key in env['featuresMap']:
+    print(key)
 ```
 
 ### Activate Feature
-The `/users/:userId/features/:key` endpoint activates the feature for a given user. In Optimizely, activation is in the context of a given user. In this case we'll provide a `userId` via a path parameter. The `userId` will be used to determine how the feature feature will be returned, if at all. Features can either be part of a Feature Test in which variations of feature variables are being measured against one another or a feature rollout, which progressively make the feature availble to a large audience.
+The `/activate?featureKey={key}` endpoint activates the feature for a given user. In Optimizely, activation is in the context of a given user to make the relative bucketing decision. In this case we'll provide a `userId` via the request body. The `userId` will be used to determine how the feature feature will be evaluated. Features can either be part of a Feature Test in which variations of feature variables are being measured against one another or a feature rollout, which progressively make the feature available to the selected audience.
```
pawels-optimizely: nit "how the feature feature" (2*feature)

docs/getting-started.md
```diff
@@ -38,35 +38,26 @@ s.headers.update({'X-Optimizely-SDK-Key': '<<YOUR-SDK-KEY>>'})
...
-feature_key = 'feature-key'
-resp = s.get('http://localhost:8080/features/{}'.format(feature_key))
-
-print(resp.json())
+for key in env['featuresMap']:
+    print(key)
 ```
 
 ### Activate Feature
-The `/users/:userId/features/:key` endpoint activates the feature for a given user. In Optimizely, activation is in the context of a given user. In this case we'll provide a `userId` via a path parameter. The `userId` will be used to determine how the feature feature will be returned, if at all. Features can either be part of a Feature Test in which variations of feature variables are being measured against one another or a feature rollout, which progressively make the feature availble to a large audience.
```
pawels-optimizely: if this is POST, then probably query parameter not needed here ?

examples/basic.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/python
+
+import json
+import requests
+import sys
+
+sdk_key = sys.argv[1]
```
pawels-optimizely: super nit, we may check for number of arguments: len(sys.argv) and exit gracefully

examples/override.py
```diff
@@ -0,0 +1,18 @@
...
+import json
+import requests
+import sys
+
+sdk_key = sys.argv[1]
+exp_key = sys.argv[2]
+var_key = sys.argv[3]
+
+s = requests.Session()
+s.headers.update({'X-Optimizely-SDK-Key': sdk_key})
```
pawels-optimizely: nice - I did not know about this way -  I usually put the header as a part of request.

pkg/routers/api_v1.go
```diff
@@ -0,0 +1,107 @@
...
+
+	if opt.maxConns > 0 {
+		// Note this is NOT a rate limiter, but a concurrency threshold
+		r.Use(chimw.Throttle(opt.maxConns))
+	}
+
+	r.Use(middleware.SetTime, opt.middleware.ClientCtx)
+	r.Use(render.SetContentType(render.ContentTypeJSON), middleware.SetRequestID)
+
+	r.With(getConfigTimer).Get("/v1/config", opt.handlers.config)
```
pawels-optimizely: I think versioning of api can be handled a little differently. Not sure if v1 is needed. if it is needed then v1 needs to be available on all the routes and can be handled globally in NewAPIRouter

api/openapi-spec/openapi.yaml
```diff
@@ -6,187 +6,55 @@ info:
...
-                   response includes the enabled status of the feature in addition to any configured variables. Any
-                   feature that is part of a feature test(s) will also send an impression event to the Optimizely
-                   analytics backend for reporting analysis.
-      tags:
-        - users
+      summary: Activate selected features and experiments for the given user.
+      operationId: activate
+      description: Activate iterates over all corresponding features
+        and experiments and returns the decision for the given user.
+        An impression event will sent to the Optimizely analytics backend
```
mjc1283:
```suggestion
        Optionally, an impression event will sent to the Optimizely analytics backend
```

pkg/routers/api_v1.go
```diff
@@ -0,0 +1,107 @@
...
+
+// APIV1Options defines the configuration parameters for Router.
+type APIV1Options struct {
+	maxConns        int
+	middleware      middleware.OptlyMiddleware
+	handlers        apiHandlers
+	metricsRegistry *metrics.Registry
+}
+
+// Define an interface to fasciliate testing
```
mjc1283:
```suggestion
// Define an interface to facilitate testing
```

pkg/routers/api_v1.go
mjc1283: At least, can we use `r.Route` to apply the `"v1"` prefix in a single place?

pkg/routers/api_v1.go
```diff
@@ -0,0 +1,107 @@
+/****************************************************************************
+ * Copyright 2019-2020, Optimizely, Inc. and contributors                        *
```
mjc1283: I think this and other new files should be 2020 only.

docs/getting-started.md
mikecdavis: The query parameter define the "what" (feature, experiment, etc) while the body is the "who" (user id and attributes).

examples/basic.py
mikecdavis: Great idea. I added the check and responded with the expected parameters. Ideally I'd use something like `argparse`, but for these examples I wanted to keep it as terse as possible.

PR #136, opened 2020-01-30
Add nil query param support.

mikecdavis:
## Summary
* Treat empty query params as nil
* Empty strings can be passed as `key=""`

Currently empty query values are treated as strings which fails the compatibility suite use cases.


mikecdavis:
@pawels-optimizely This is stop gap, when we move to putting attributes in the request body, this type coercion can be deprecated.

PR #137, opened 2020-01-31
feat[OASIS-5898]: Handlers for client credential flow (2-legged OAuth)

mjc1283:
## Summary
Implements handlers for issuing access tokens for the API and admin services.
- Supporting changes are added in config
- A new package `jwtauth` contains auth-related logic
- Several TODOs remain for follow-up PRs, including connecting these handlers to routers


pkg/handlers/oauth.go
```diff
@@ -0,0 +1,156 @@
...
+	clientCreds, ok := h.ClientCredentials[reqBody.ClientID]
+	// TODO: hash client secret and match secret hash
+	if !ok || !jwtauth.MatchClientSecret(reqBody.ClientSecret, clientCreds.Secret) {
+		return nil, http.StatusForbidden, errors.New("invalid client_id or client_secret")
+	}
+	return &clientCreds, http.StatusOK, nil
+}
+
+// GetAPIAccessToken returns a JWT access token for the API service
+func (h *OAuthHandler) GetAPIAccessToken(w http.ResponseWriter, r *http.Request) {
```
mikecdavis: Nit, but should these be `Create` as opposed to `Get`? `Get` in a handler sounds like HTTP GET :P

mikecdavis: Super nit, but can we prefix these with `Create` as oppose to `Get`? Get sounds like HTTP GET in the context of a handler.

pkg/handlers/oauth.go
```diff
@@ -0,0 +1,156 @@
...
+
+	clientCreds, httpCode, e := h.verifyClientCredentials(r)
```
mikecdavis: can we use `err` as opposed to `e`?

pkg/handlers/oauth_test.go
```diff
@@ -0,0 +1,256 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,        *
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *
+ * See the License for the specific language governing permissions and      *
+ * limitations under the License.                                           *
+ ***************************************************************************/
+
+// Package handlers //
+package handlers
+
+import (
```
pawels-optimizely: nit, organize imports

pkg/handlers/oauth.go
pawels-optimizely: IssueAPIAccessToken may be better

pkg/handlers/oauth.go
```diff
@@ -0,0 +1,156 @@
...
+		middleware.GetLogger(r).Error().Err(err).Msg("Calling jwt BuildAPIAccessToken")
+		RenderError(err, http.StatusInternalServerError, w, r)
+		return
+	}
+
+	renderAccessTokenResponse(w, r, accessToken, expires)
+}
+
+// GetAdminAccessToken returns a JWT access token for the Admin service
+func (h *OAuthHandler) GetAdminAccessToken(w http.ResponseWriter, r *http.Request) {
```
pawels-optimizely: same as before

pkg/jwtauth/jwtauth.go
```diff
@@ -0,0 +1,67 @@
...
+ *                                                                          *
+ * Unless required by applicable law or agreed to in writing, software      *
+ * distributed under the License is distributed on an "AS IS" BASIS,        *
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *
+ * See the License for the specific language governing permissions and      *
+ * limitations under the License.                                           *
+ ***************************************************************************/
+
+// Package jwtauth contains JWT-related helpers
+package jwtauth
```
mikecdavis: can we have this as `jwt`?

pkg/jwtauth/jwtauth.go
```diff
@@ -0,0 +1,67 @@
...
+	})
+	tokenString, err = token.SignedString(key)
+	if err != nil {
+		return "", 0, fmt.Errorf("error building Admin access token: %w", err)
+	}
+	return tokenString, expires, nil
+}
+
+// MatchClientSecret compares secret keys
+func MatchClientSecret(reqSecretStr string, configSecret []byte) bool {
```
mikecdavis: Nit, `ValidateClientSecret`?

pkg/jwtauth/jwtauth.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"crypto/subtle"
+	"fmt"
+	"time"
+
+	"github.com/dgrijalva/jwt-go"
+)
+
+// BuildAPIAccessToken returns a token for accessing the API service using the argument SDK key and TTL. It also returns the expiration timestamp.
+func BuildAPIAccessToken(sdkKey string, ttl time.Duration, key []byte) (tokenString string, expires int64, err error) {
```
mikecdavis: We shouldn't need to return the `expires` since we passed it in as in input parameter. Expires should be the TTL not the timestamp.

pkg/handlers/oauth.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+type tokenResponse struct {
+	AccessToken string `json:"access_token"`
+	TokenType   string `json:"token_type"`
+	ExpiresIn   int64  `json:"expires_in"`
+}
+
+func renderAccessTokenResponse(w http.ResponseWriter, r *http.Request, accessToken string, expires int64) {
+	// TODO: expires_in should be in seconds, per https://tools.ietf.org/html/rfc6749#section-5.1
```
mikecdavis: Should just be `clientCreds.TTL` in seconds.

config/config.go
```diff
@@ -115,3 +127,16 @@ type WebhookProject struct {
...
+// OAuthClientCredentials are used for issuing access tokens
+type OAuthClientCredentials struct {
+	ID     string `yaml:"id"`
+	Secret string `yaml:"secret"`
+}
+
+// ServiceAuthConfig holds the authentication configuration for a particular service
+type ServiceAuthConfig struct {
+	Clients    []OAuthClientCredentials `yaml:"clients"`
+	HMACSecret string                   `yaml:"hmacSecret"`
```
mikecdavis: must include ``json:"-"`` so that these are exposed when marshalled to json via the `/config` endpoint

pkg/jwtauth/jwtauth.go
pawels-optimizely: we use jwt package already (from dgrijalva), that's why I think we should have a different name here if possible.

mjc1283: Initially I named it this way because It also has `MatchClientSecret` which isn't _really_ JWT-related, but on second thought, I guess it's fine to use `jwt`.

PR #138, opened 2020-02-01
feat: token acceptor for admin and api

pawels-optimizely:
## Summary
- adding Authorize middleware for accepting a token when auth is enabled 
- for now enabled/disabled auth is determined by the presence of auth object in config file

### DIFF
--- a/pkg/middleware/auth.go
+++ b/pkg/middleware/auth.go
@@ -24,6 +24,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/optimizely/agent/config"
+
 	"github.com/dgrijalva/jwt-go"
 )
 
@@ -48,8 +50,6 @@ func (NoAuth) CheckToken(string) (*jwt.Token, error) {
 // Auth is the middleware for all REST API's
 type Auth struct {
 	Verifier
-
-	checkClaims map[string]struct{}
 }
 
 // Verifier checks token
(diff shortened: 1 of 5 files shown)


pkg/middleware/auth.go
```diff
@@ -0,0 +1,175 @@
...
+
+	var token string
+
+	if values, ok := r.Header["Auth"]; ok && len(values) > 0 {
+		token = values[0]
+	}
+
+	if values, ok := r.Header["Jwt"]; ok && len(values) > 0 {
+		token = values[0]
+	}
```
mikecdavis: Are these standard header keys? Can we move to configuration as opposed to sniffing these out?

pkg/middleware/auth.go
```diff
@@ -0,0 +1,175 @@
...
+	return a.CheckToken(token)
+
+}
+
+func (a Auth) enabled() bool {
+	if _, ok := a.Verifier.(NoAuth); ok {
+		return false
+	}
+	return true
+}
```
mikecdavis: If we move to the claims checks to independent middleware then this is not needed.

pkg/middleware/auth.go
```diff
@@ -0,0 +1,175 @@
...
+
+// Authorize is middleware for auth
+func (a Auth) Authorize(next http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+
+		tk, err := a.verify(r)
+
+		if err != nil {
+			http.Error(w, fmt.Sprintf(`{"error": "unauthorized, "reason": "%v"}`, err), http.StatusUnauthorized)
```
mikecdavis: We should be using the `RenderError` helper method since it uses `render.JSON` which sanitizes the output to prevent JS injection. I know there's no injection possible here, but making sure everything goes through `render.JSON` is to avoid potential security risks.

pkg/middleware/auth.go
```diff
@@ -0,0 +1,175 @@
...
+			return
+		}
+
+		if a.enabled() {
+			claims := tk.Claims.(jwt.MapClaims)
+
+			for key := range a.checkClaims {
```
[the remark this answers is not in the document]
mjc1283: +1 to separating the signature & expiry validation from the claims validation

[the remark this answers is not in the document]
pawels-optimizely: this is already done, CheckToken checks for a valid token and expiry. I don't need to really check for exp here, but I am doing that anyway to make sure. If the token is expired, CheckToken will return an error.

pkg/middleware/auth.go
```diff
@@ -0,0 +1,175 @@
...
+
+				value, ok := claims[key]
+				if !ok {
+					http.Error(w, fmt.Sprintf(`{"error": "unauthorized, "reason": "%s key not in claims"}`, key), http.StatusUnauthorized)
+					return
+				}
+
+				switch key {
+				case "exp":
```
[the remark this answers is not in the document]
pawels-optimizely: I can move it to a global check to make sure

pkg/middleware/auth.go
pawels-optimizely: I put here most popular and common prefixes for token checks

PR #140, opened 2020-02-11
Add override handler.

mikecdavis:
## Summary
* Add override handler

To keep the scope small I'm breaking out the changes in PR #133 into several smaller PRs.

### DIFF
--- a/pkg/handlers/override.go
+++ b/pkg/handlers/override.go
@@ -46,6 +46,11 @@ func Override(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if body.UserID == "" {
+		RenderError(errors.New("userId cannot be empty"), http.StatusBadRequest, w, r)
+		return
+	}
+
 	experimentKey := body.ExperimentKey
 	if experimentKey == "" {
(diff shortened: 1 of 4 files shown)


pkg/handlers/override.go
```diff
@@ -0,0 +1,76 @@
...
+		if err != nil {
+			RenderError(err, http.StatusInternalServerError, w, r)
+			return
+		}
+		w.WriteHeader(http.StatusNoContent)
+		return
+	}
+
+	logger.Debug().Str("experimentKey", experimentKey).Str("variationKey", body.VariationKey).Msg("setting override")
+	wasSet, err := optlyClient.SetForcedVariation(experimentKey, body.UserID, body.VariationKey)
```
mjc1283: So we aren't using the middleware for obtaining user context anymore? Should we be returning 400 when userID is empty?

mikecdavis: Good point I will fix that. The middleware concept isn't advantageous with these new routes since the models aren't shared between handlers like they were before. The only middleware that is still required is to fetch the Optimizely client.

PR #143, opened 2020-02-11
Add activate handler.

mikecdavis:
## Summary
* Add activate handler

To keep the scope small I'm breaking out the changes in PR #133 into several smaller PRs.

### DIFF
--- a/pkg/handlers/activate.go
+++ b/pkg/handlers/activate.go
@@ -82,8 +82,8 @@ func Activate(w http.ResponseWriter, r *http.Request) {
 		featureSet[key] = struct{}{}
 	}
 
-	logger.Debug().Msg("iterate over all experiment decisions")
 	for key := range experimentSet {
+		logger.Debug().Str("experimentKey", key).Msg("fetching experiment decision")
 		e, ok := oConf.ExperimentsMap[key]
 		if !ok {
 			RenderError(fmt.Errorf("experimentKey not-found"), http.StatusNotFound, w, r)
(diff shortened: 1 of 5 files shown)


pkg/handlers/activate.go
```diff
@@ -0,0 +1,132 @@
...
+			return
+		}
+
+		d, err := optlyClient.ActivateExperiment(&e, uc, disableTracking)
+		if err != nil {
+			RenderError(err, http.StatusInternalServerError, w, r)
+			return
+		}
+
+		// TODO implement enabled filtering
```
mjc1283: JIRA?

pkg/handlers/activate.go
```diff
@@ -0,0 +1,132 @@
...
+	for _, key := range query["experimentKey"] {
+		experimentSet[key] = struct{}{}
+	}
+
+	// Add explicit features
+	for _, key := range query["featureKey"] {
+		featureSet[key] = struct{}{}
+	}
+
+	logger.Debug().Msg("iterate over all experiment decisions")
```
mjc1283: This doesn't seem very useful. Maybe better to include the keys that will be decided on?

pkg/handlers/activate.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+
+		// TODO implement enabled filtering
+		decisions = append(decisions, d)
+	}
+
+	render.JSON(w, r, decisions)
+}
+
+func getUserContext(r *http.Request) (entities.UserContext, error) {
```
mjc1283: Maybe this function should be reused in the track or override handlers.

pkg/handlers/activate.go
mikecdavis: I adding the functionality.

pkg/handlers/activate.go
mikecdavis: moved to a debug inside the loop

PR #147, opened 2020-02-13
feat: Add blacklist for crypto cyphers

pawels-optimizely:
## Summary
- add a configurable blacklist for ciphers
- the implementation removes the cipher from our default config if that cipher is blacklisted
- blacklist is provided in configuration as a list of cipher hexadecimal values

### DIFF
--- a/cmd/testdata/default.yaml
+++ b/cmd/testdata/default.yaml
@@ -10,6 +10,9 @@ server:
   writeTimeout: 10s
   keyfile: "keyfile"
   certfile: "certfile"
+  disabledciphers:
+    - 0xc00a
+    - 0xc011
 log:
   pretty: true
   level: debug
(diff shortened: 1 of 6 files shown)


cmd/testdata/default.yaml
```diff
@@ -10,6 +10,9 @@ server:
   writeTimeout: 10s
   keyfile: "keyfile"
   certfile: "certfile"
+  disabledciphers:
+    - 0xc00a
+    - 0xc011
```
mjc1283: Is it inconvenient for the user to make them figure out the hex value? Should it be a semantic string?

pawels-optimizely:
> Talked offline with @pawels-optimizely - the hex value is part of the standard, so I think it's OK to use directly.

That's correct. A user can go to this official page https://www.iana.org/assignments/tls-parameters/tls-parameters.xml#tls-parameters-4 and get the hexadecimal value corresponded to the cypher.  

As far as documentation goes, nothing is documented for security section as of now, and there is a separate task for it. We will add documentation for all security topics as they are finished.

cmd/testdata/default.yaml
xarses: Given other configuration formats, I'd expect people would expect string format.

Here is an example from a white-list configuration [sauce](https://jenda.hrach.eu/f2/cipherlist.html)
> ECDHE-RSA-AES256-GCM-SHA512:DHE-RSA-AES256-GCM-SHA512:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-SHA384

PR #150, opened 2020-02-20
feat: Disable /notifications and /overrides by default

pawels-optimizely:
## Summary
- disable /notifications and /overrides by default
- allow them to be enabled via configuration

### DIFF
--- a/cmd/testdata/default.yaml
+++ b/cmd/testdata/default.yaml
@@ -24,6 +24,8 @@ admin:
 api:
   maxconns: 100
   port: "3000"
+  enablenotifications: true
+  enableoverrides: true
   auth:
     ttl: 30m
     hmacSecret: "abcd"
--- a/pkg/handlers/notification_disabled.go
+++ b/pkg/handlers/notification_disabled.go
@@ -1,0 +1,37 @@
+/****************************************************************************
+ * Copyright 2020, Optimizely, Inc. and contributors                        *
+ *                                                                          *
+ * Licensed under the Apache License, Version 2.0 (the "License");          *
+ * you may not use this file except in compliance with the License.         *
+ * You may obtain a copy of the License at                                  *
+ *                                                                          *
+ *    http://www.apache.org/licenses/LICENSE-2.0                            *
+ *                                                                          *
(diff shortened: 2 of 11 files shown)


pkg/routers/api_v1_test.go
```diff
@@ -72,6 +72,7 @@ func (suite *APIV1TestSuite) SetupTest() {
 		middleware:      &MockOptlyMiddleware{},
 		handlers:        MockHandlers{},
 		metricsRegistry: metricsRegistry,
+		enableOverrides: true,
```
mjc1283: I think we need a new test in this file, covering changes to `routers/api_v1.go`.

pkg/handlers/notification_disabled.go
```diff
@@ -0,0 +1,37 @@
...
+
+// Package handlers //
+package handlers
+
+import (
+	"net/http"
+)
+
+// DisableNotificationHandler for disabled notifications
+type DisableNotificationHandler struct {
```
mikeng13: nit: I'd name this `DisabledNotificationHandler` notice `Disable` vs `Disabled`

pkg/handlers/user_overrides_disabled.go
```diff
@@ -0,0 +1,36 @@
...
+
+// Package handlers //
+package handlers
+
+import (
+	"net/http"
+)
+
+// DisableUserOverrideHandler implements the disabled overrides
+type DisableUserOverrideHandler struct{}
```
mikeng13: Same comment on `Disabled`

cmd/testdata/default.yaml
```diff
@@ -24,6 +24,8 @@ admin:
 api:
   maxconns: 100
   port: "3000"
+  enablenotifications: true
+  enableoverrides: true
```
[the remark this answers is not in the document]
pawels-optimizely: > Suggest `enable_notifications`, `enable_overrides`

I know, I just tried to follow majority patterns, but it's still not consistent. We have a camel case: flushInterval or batchSize, all small case: sdkkeys or readtimeout. Introducing a snake case will not look nice. I think we need to clean up the definitions and go with either a camel case or snake case for all.

PR #153, opened 2020-02-24
Handle invalid filter types.

mikecdavis:
## Summary
* Return error if `?type=` parameter is not supported

Current behavior is to ignore the unknown type filter.

### DIFF
--- a/pkg/handlers/activate.go
+++ b/pkg/handlers/activate.go
@@ -21,11 +21,10 @@ import (
 	"fmt"
 	"net/http"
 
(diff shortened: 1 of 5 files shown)


pkg/handlers/activate.go
```diff
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/optimizely/go-sdk/pkg/config"
```
pawels-optimizely: nit, group with github.com/optimizely/go-sdk/pkg/entities

PR #159, opened 2020-02-25
Throw 404 if eventKey not part of the project

mikecdavis:
## Summary
* Create `SycnedConfigManager` interface for OptlyClient
* Check the OptlyClient config manager for event key prior to calling track

This is a bit of a workaround since the native SDK swallows tracking errors related to event key existence and events are not part of the `OptimizelyConfig` object. We could have opened up the SDK a bit, but that would have amounted to a breaking change if we started to propagate errors (which is arguably the right approach). That being said, for track in particular I want to set us up to remove the event creation constraint within the application in which case we'd simply remove this code as 404 will no longer be applicable.

### DIFF
--- a/pkg/handlers/track.go
+++ b/pkg/handlers/track.go
@@ -73,7 +73,7 @@ func TrackEvent(w http.ResponseWriter, r *http.Request) {
 
 	err = optlyClient.Track(eventKey, uc, body.EventTags)
 	if err != nil {
-		RenderError(err, http.StatusBadRequest, w, r)
+		RenderError(err, http.StatusInternalServerError, w, r)
 		return
 	}
 
--- a/pkg/handlers/track.go
+++ b/pkg/handlers/track.go
@@ -60,13 +60,23 @@ func TrackEvent(w http.ResponseWriter, r *http.Request) {
 		Attributes: body.UserAttributes,
 	}
 
-	err = optlyClient.Track(eventKey, uc, body.EventTags)
-
+	pc, err := optlyClient.ConfigManager.GetConfig()
 	if err != nil {
-		middleware.GetLogger(r).Error().Err(err).Str("eventKey", eventKey).Msg("error tracking event")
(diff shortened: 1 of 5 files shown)


pkg/optimizely/interface.go
```diff
@@ -17,7 +17,18 @@
...
+	optimizelyconfig "github.com/optimizely/go-sdk/pkg/config"
+)
+
 // Cache defines a basic interface for retrieving an instance of the OptlyClient keyed off of the SDK Key
 type Cache interface {
 	GetClient(sdkKey string) (*OptlyClient, error)
 }
+
+// SyncedConfigManager has the basic ConfigManager methods plus the SyncConfig method to trigger immediate updates
+type SyncedConfigManager interface {
```
pawels-optimizely: is this SyncedConfigManager created only to help with testing ?

pkg/handlers/track.go
```diff
@@ -60,13 +60,23 @@ func TrackEvent(w http.ResponseWriter, r *http.Request) {
...
 		RenderError(err, http.StatusNotFound, w, r)
 		return
 	}
+
+	err = optlyClient.Track(eventKey, uc, body.EventTags)
+	if err != nil {
+		RenderError(err, http.StatusInternalServerError, w, r)
+		return
+	}
+
```
mjc1283: Nit: this logic could be inside the `optlyClient`, and it could return well known errors so the caller can decide whether to 500 or 404. Then the handler isn't messing with the config manager or the project config.

pkg/optimizely/interface.go
mikecdavis: Yes. The `SyncConfig` method only exists on the `PollingConfigManager` so we've been using that explicit reference, this abstracts that out a bit for testing and defines only the methods we're actually using.

pkg/handlers/track.go
mikecdavis: You're right, this should be part of `OptlyClient`.

PR #160, opened 2020-02-25
feat: Connect access token creation handler and auth middleware in V1 API

mjc1283:
## Summary
This makes the existing OAuth functionality available in the V1 API service:
- Exposed token issuer endpoint at `POST /v1/oauth/token`
- Connected token authorization middleware to `v1/config`, `v1/activate`, `v1/track`, and `v1/override`
## Test Plan
- Manually tested
- Updated V1 router unit tests
- Existing tests for old API and Admin services should pass
## JIRA
https://optimizely.atlassian.net/browse/OASIS-6028

### DIFF
--- a/pkg/routers/api_v1.go
+++ b/pkg/routers/api_v1.go
@@ -38,6 +38,7 @@ type APIV1Options struct {
 	middleware      middleware.OptlyMiddleware
 	handlers        apiHandlers
 	metricsRegistry *metrics.Registry
+	oAuthHandler    *handlers.OAuthHandler
 	oAuthMiddleware middleware.Auth
 }
 
@@ -47,16 +48,9 @@ type apiHandlers interface {
 	activate(w http.ResponseWriter, r *http.Request)
 	trackEvent(w http.ResponseWriter, r *http.Request)
 	override(w http.ResponseWriter, r *http.Request)
-	createAccessToken(w http.ResponseWriter, r *http.Request)
 }
 
-type defaultHandlers struct {
-	oAuthHandler *handlers.OAuthHandler
-}
-
-func newDefaultHandlers(oAuthHandler *handlers.OAuthHandler) *defaultHandlers {
-	return &defaultHandlers{oAuthHandler}
-}
+type defaultHandlers struct{}
 
 func (d defaultHandlers) config(w http.ResponseWriter, r *http.Request) {
 	handlers.OptimizelyConfig(w, r)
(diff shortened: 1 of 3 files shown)


pkg/routers/api_v1.go
```diff
@@ -66,13 +68,23 @@ func (d defaultHandlers) override(w http.ResponseWriter, r *http.Request) {
...
 }
 
+type apiOAuthHandler interface {
+	CreateAPIAccessToken(w http.ResponseWriter, r *http.Request)
+}
+
+type apiOAuthMiddleware interface {
+	AuthorizeAPI(next http.Handler) http.Handler
+}
+
```
mikecdavis: FWIW, I'm planning on removing these interfaces. I think they are getting unruly and not adding much benefit. I think this is OK for now, but just a heads up that a refactor is on the horizon :)

pkg/routers/api_v1.go
```diff
@@ -110,9 +123,10 @@ func WithAPIV1Router(opt *APIV1Options, r chi.Router) {
...
 	r.Route("/v1", func(r chi.Router) {
-		r.With(getConfigTimer).Get("/config", opt.handlers.config)
-		r.With(activateTimer).Post("/activate", opt.handlers.activate)
-		r.With(trackTimer).Post("/track", opt.handlers.trackEvent)
-		r.With(overrideTimer).Post("/override", overrideHandler)
+		r.With(getConfigTimer, opt.oAuthMiddleware.AuthorizeAPI).Get("/config", opt.handlers.config)
+		r.With(activateTimer, opt.oAuthMiddleware.AuthorizeAPI).Post("/activate", opt.handlers.activate)
+		r.With(trackTimer, opt.oAuthMiddleware.AuthorizeAPI).Post("/track", opt.handlers.trackEvent)
+		r.With(overrideTimer, opt.oAuthMiddleware.AuthorizeAPI).Post("/override", overrideHandler)
+		r.With(createAccesstokenTimer).Post("/oauth/token", opt.oAuthHandler.CreateAPIAccessToken)
```
mikecdavis: Hm, I'm curious if this even makes sense under `/v1`? This API versioning is for the Agent API, the auth flow is more or less prescribed by the OAuth 2.0 spec and not under the same versioning scheme as our API. Thoughts?

mjc1283: The SDK key header is required, and included in the claims. Does it make sense to say, in theory, this may need to be versioned alongside the Agent API, because we might remove or change the SDK key header?

mjc1283: The SDK key changing seems very unlikely so I would agree with moving this out of `/v1`. Probably its path should include `api` somewhere to indicate that it gives API access, as opposed to admin access.

PR #162, opened 2020-02-27
feat: Add support for JWKS URI

pawels-optimizely:
## Summary
- adding support for JWKS to validate the token signature
- token claims must include `kid` field - this is the standard way of validating a token against the key set returning by jwks url
- jwks_url supports `http://`, `https://`, and `file://` as the valid url scheme


pkg/middleware/auth.go
```diff
@@ -90,6 +91,52 @@ func (c JWTVerifier) CheckToken(token string) (*jwt.Token, error) {
...
+
+// NewJWTVerifierURL creates JWTVerifierURL with JWKS URL
+func NewJWTVerifierURL(jwksURL string) JWTVerifierURL {
+	return JWTVerifierURL{jwksURL: jwksURL, parser: new(jwt.Parser)}
+}
+
+// CheckToken checks the token, validates against JWKS and returns it if it's valid
+func (c JWTVerifierURL) CheckToken(token string) (*jwt.Token, error) {
+	if token == "" {
+		return nil, errors.New("invalid token")
```
mjc1283: Errors returned from `CheckToken` end up rendered into the response, and the response status is 401. Are you sure all these errors should be displayed to the user, and should be 401? For example I think if the fetch fails, perhaps 500 or 503 is better. This comment also applies to the `CheckToken` method of `JWTVerifier`.

pawels-optimizely: @mjc1283 and I talked about it, 401 is fine, we will just change the message to "empty token"

pkg/middleware/auth_test.go
```diff
@@ -105,6 +106,40 @@ func (suite *AuthTestSuite) TestAuthInvalidCheckToken() {
...
+	dir, _ := os.Getwd()
+	authConfig := &config.ServiceAuthConfig{
+		Clients:    make([]config.OAuthClientCredentials, 0),
+		HMACSecret: suite.signature,
+		TTL:        0,
+		JwksURL:    "file://" + dir + "/testdata/jwks_url.txt",
```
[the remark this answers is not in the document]
pawels-optimizely: we also talked online, that's actually the best test because it tests the whole flow of signing. 
I cannot mock jwk.Fetch() itself, I would have to mock the whole object, which is not the point here.

pkg/middleware/auth.go
mikecdavis: Agreed that the code should reflect user vs server error.

pkg/middleware/auth.go
```diff
@@ -90,6 +91,52 @@ func (c JWTVerifier) CheckToken(token string) (*jwt.Token, error) {
...
+	}
+
+	tk, err := c.parser.Parse(token, func(token *jwt.Token) (interface{}, error) {
```
mikecdavis: `func(token *jwt.Token) (interface{}, error)` is really the only differnce between this and the the static JWTVerifier. Consider changing the interface to reflect that abstraction.

config/config.go
```diff
@@ -34,6 +34,7 @@ func NewDefaultConfig() *AgentConfig {
 				Clients:    make([]OAuthClientCredentials, 0),
 				HMACSecret: "",
 				TTL:        0,
+				JwksURL:    "",
```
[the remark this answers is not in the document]
pawels-optimizely: probably not, we just put  them here  for consistency every time we add something new.

pkg/middleware/auth.go
```diff
@@ -90,6 +101,95 @@ func (c JWTVerifier) CheckToken(token string) (*jwt.Token, error) {
...
+
+	http.DefaultClient = &http.Client{Timeout: 10 * time.Second}
+	jwtVerifierURL := JWTVerifierURL{jwksURL: jwksURL, parser: new(jwt.Parser)}
+	err := jwtVerifierURL.updateKeySet()
+
+	if err != nil {
+		return &JWTVerifierURL{}
+	}
+
+	go jwtVerifierURL.startTicker(updateInterval)
```
mjc1283: Should `NewJWTVerifierURL` accept a context, and cancel its ticker/ in-flight requests when the context is cancelled?

pkg/middleware/auth.go
```diff
@@ -90,6 +101,95 @@ func (c JWTVerifier) CheckToken(token string) (*jwt.Token, error) {
...
+	c.jwksLock.RLock()
+	defer c.jwksLock.RUnlock()
+	return c.jwksKeys
+
+}
+
+// NewJWTVerifierURL creates JWTVerifierURL with JWKS URL
+func NewJWTVerifierURL(jwksURL string, updateInterval time.Duration) *JWTVerifierURL {
+
+	http.DefaultClient = &http.Client{Timeout: 10 * time.Second}
```
mjc1283: Is this going to affect the timeout or other settings on other clients?

pkg/middleware/auth.go
```diff
@@ -90,6 +101,95 @@ func (c JWTVerifier) CheckToken(token string) (*jwt.Token, error) {
...
+	jwtVerifierURL := JWTVerifierURL{jwksURL: jwksURL, parser: new(jwt.Parser)}
+	err := jwtVerifierURL.updateKeySet()
+
+	if err != nil {
+		return &JWTVerifierURL{}
```
mjc1283: It seems dangerous to use a `JWTVerifierURL` with uninitialized `parser`, we call `c.parser.Parse` in `CheckToken` without a nil check. Is there a test case where `updateKeySet` returns an error on this first call?

pkg/middleware/auth.go
```diff
@@ -181,6 +281,23 @@ func (a Auth) AuthorizeAPI(next http.Handler) http.Handler {
...
+		log.Warn().Msg("HMAC Secret will be ignored, JWKS URL will be used for token validation")
+	}
+
+	if authConfig.JwksURL != "" {
+		if authConfig.JwksUpdateInterval <= 0 {
+			log.Error().Msg("JwksUpdateInterval must be set")
+			return Auth{Verifier: BadAuth{}}
+		}
+		verifier := NewJWTVerifierURL(authConfig.JwksURL, authConfig.JwksUpdateInterval)
+		if verifier.jwksKeys == nil {
```
mjc1283: Instead of `verifier.jwksKeys == nil`, perhaps: `verifier == nil` or `verifier, err := NewJWTVerifierURL(..); err != nil` would be clearer.

pkg/middleware/auth.go
pawels-optimizely: I don't think it's needed. it's just the updater, and it will exit when the main thread exits. We don't have to wait for the complete update.

pkg/middleware/auth.go
pawels-optimizely: I don't know if it's worthwhile to introduce another parameter for this timeout.

pkg/middleware/auth.go
pawels-optimizely: if the error is not null then it will not even hit CheckToken method. The program will exit and the ticker won't start.

pkg/middleware/auth.go
pawels-optimizely: verifier is the interface, I was returning everything by value for other cases. NewJWTVerifierURL needs to return by pointer because of locking logic. I can make other cases consistent, return by pointer, and then remove BadAuth.
`NewAuth(authConfig *config.ServiceAuthConfig) Auth` will need to become `NewAuth(authConfig *config.ServiceAuthConfig) *Auth` .

pkg/middleware/auth.go
```diff
@@ -47,6 +50,14 @@ func (NoAuth) CheckToken(string) (*jwt.Token, error) {
 	return nil, nil
 }
 
+// BadAuth is bad auth, it shuts down the server
+type BadAuth struct{}
```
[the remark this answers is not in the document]
pawels-optimizely: errors are not the best option either. I think I can return Auth by pointer and then check for null pointer.

pkg/middleware/auth_test.go
```diff
@@ -203,6 +345,81 @@ func (suite *AuthTestSuite) TestAuthAuthorizeAdminTokenAuthorizationValidClaimsE
...
+
+	verifier.jwksLock.Lock()
+	verifier.jwksURL = validJwksURL
+	verifier.jwksLock.Unlock()
+
+	token, err = auth.CheckToken(tk)
+	assert.Nil(t, token) // still bad - polling every minute, still using fake_url
+	assert.Error(t, err)
+
+	<-time.After(1500 * time.Millisecond)
```
mjc1283: Any way to make this faster than 1500 ms?

config/config.go
```diff
@@ -31,17 +31,21 @@ func NewDefaultConfig() *AgentConfig {
...
 		API: APIConfig{
 			Auth: ServiceAuthConfig{
-				Clients:     make([]OAuthClientCredentials, 0),
-				HMACSecrets: make([]string, 0),
-				TTL:         0,
+				Clients:            make([]OAuthClientCredentials, 0),
+				HMACSecrets:        make([]string, 0),
+				TTL:                0,
+				JwksURL:            "",
+				JwksUpdateInterval: 0,
```
mjc1283: Could you add JWKS update interval to the table of config parameters in the main README?

pkg/middleware/auth.go
```diff
@@ -99,6 +102,95 @@ func (c JWTVerifier) CheckToken(token string) (*jwt.Token, error) {
...
+// NewJWTVerifierURL creates JWTVerifierURL with JWKS URL
+func NewJWTVerifierURL(jwksURL string, updateInterval time.Duration) *JWTVerifierURL {
+
+	http.DefaultClient = &http.Client{Timeout: 10 * time.Second}
+	jwtVerifierURL := JWTVerifierURL{jwksURL: jwksURL, parser: new(jwt.Parser)}
+	err := jwtVerifierURL.updateKeySet()
+
+	if err != nil {
+		return nil
+	}
```
mikecdavis: Why not return the error? Right now it's getting swallowed.

pkg/routers/admin.go
```diff
@@ -34,6 +34,10 @@ func NewAdminRouter(conf config.AgentConfig) http.Handler {
 
 	authProvider := middleware.NewAuth(&conf.Admin.Auth)
 
+	if authProvider == nil {
+		return nil
+	}
```
mikecdavis: Similar here, propagating the error feels more idiomatic.

pkg/middleware/auth.go
mikecdavis: If we want to return `nil` and not the error, then we should log the error here.
```
log.Error().Err(err).Msg("problem with getting initial JWKS key set")
```

pkg/middleware/auth.go
```diff
@@ -188,12 +280,29 @@ func (a Auth) AuthorizeAPI(next http.Handler) http.Handler {
...
+
+	if authConfig.JwksURL != "" {
+		if authConfig.JwksUpdateInterval <= 0 {
+			log.Error().Msg("JwksUpdateInterval must be set")
+			return nil
+		}
+		verifier := NewJWTVerifierURL(authConfig.JwksURL, authConfig.JwksUpdateInterval)
+		if verifier == nil {
+			log.Error().Msg("problem with getting initial JWKS key set")
```
mikecdavis: This is leaking a bit as we're inferring why the verifier returned `nil`. The best we can say here is:
```
log.Error().Msg("unable to construct NewJWTVerifierURL")
```

pkg/routers/admin.go
mikecdavis: Should log here as well:
```
log.Error().Msg("unable to initialize auth middleware.")
```

pkg/routers/api.go
```diff
@@ -43,13 +43,16 @@ type APIOptions struct {
...
+	oAuthMiddleware  *middleware.Auth
 }
 
 // NewDefaultAPIRouter creates a new router with the default backing optimizely.Cache
 func NewDefaultAPIRouter(optlyCache optimizely.Cache, conf config.APIConfig, metricsRegistry *metrics.Registry) http.Handler {
 
 	authProvider := middleware.NewAuth(&conf.Auth)
+	if authProvider == nil {
+		return nil
+	}
```
mikecdavis: Same here, we should log why we're returning `nil`:
```
log.Error().Msg("unable to initialize auth middleware.")
```

cmd/main.go
```diff
@@ -118,10 +118,18 @@ func main() {
 		cancel()
 	}()
 
+	apiRouter := routers.NewDefaultAPIRouter(optlyCache, conf.API, agentMetricsRegistry)
+	adminRouter := routers.NewAdminRouter(*conf)
+
+	if apiRouter == nil || adminRouter == nil {
+		log.Error().Msg("failed to start router, exiting")
+		cancel()
+	}
```
mikecdavis: We're loosing the granularity about which router is failing. We probably don't need to assign explicit variables since we can do the `nil` check within `NewServer` which should log with all the important details.

pkg/middleware/auth.go
```diff
@@ -99,6 +102,95 @@ func (c JWTVerifier) CheckToken(token string) (*jwt.Token, error) {
...
+	jwksKeys *jwk.Set
+	jwksLock sync.RWMutex
+}
+
+func (c *JWTVerifierURL) startTicker(ticker time.Duration) {
+
+	for range time.Tick(ticker) {
+		err := c.updateKeySet()
+		if err != nil {
+			log.Warn().Msg("unable to update JWKS key set")
```
mikecdavis: Should add the error to the log message:
```
log.Warn().Err(err).Msg("unable to update JWKS key set")
```

pkg/middleware/auth.go
mikecdavis: From [the docs](https://golang.org/pkg/time/#Tick)

> Tick is a convenience wrapper for NewTicker providing access to the ticking channel only. While Tick is useful for clients that have no need to shut down the Ticker, be aware that without a way to shut it down the underlying Ticker cannot be recovered by the garbage collector; it "leaks". Unlike NewTicker, Tick will return nil if d <= 0.

pkg/middleware/auth.go
pawels-optimizely: I think this one should stay as a warning, because we don't terminate the program

pawels-optimizely:
> We need to be careful about swallowing errors and providing clear messages. This might be less of an issue if we were propagating the offending error, but I'm ok with nil for now.

ok, I will add the error to the constructor

pkg/middleware/auth.go
mikecdavis: this is still a warning (e.g. `.Warn()`), we're just adding the error to the output. Note `.Err(err)` is **not** a log level but a way to append the error message. Basically shorthand for `.Str("error", err.Error())`.

pawels-optimizely: ok, I missed that

cmd/main.go
mikecdavis: This comment has not been addressed.

pkg/middleware/auth.go
```diff
@@ -99,6 +102,99 @@ func (c JWTVerifier) CheckToken(token string) (*jwt.Token, error) {
...
+
+// NewJWTVerifierURL creates JWTVerifierURL with JWKS URL
+func NewJWTVerifierURL(jwksURL string, updateInterval time.Duration) *JWTVerifierURL {
+
+	http.DefaultClient = &http.Client{Timeout: 10 * time.Second}
+	jwtVerifierURL := JWTVerifierURL{jwksURL: jwksURL, parser: new(jwt.Parser)}
+	err := jwtVerifierURL.updateKeySet()
+
+	if err != nil {
+		log.Error().Msg("unable to construct NewJWTVerifierURL")
```
mikecdavis: We're swallowing the error here. Should be:
```go
log.Error().Err(err).Msg("unable to fetch key set")
```

pkg/middleware/auth.go
```diff
@@ -99,6 +102,99 @@ func (c JWTVerifier) CheckToken(token string) (*jwt.Token, error) {
...
+
+// NewJWTVerifierURL creates JWTVerifierURL with JWKS URL
+func NewJWTVerifierURL(jwksURL string, updateInterval time.Duration) *JWTVerifierURL {
+
+	http.DefaultClient = &http.Client{Timeout: 10 * time.Second}
+	jwtVerifierURL := JWTVerifierURL{jwksURL: jwksURL, parser: new(jwt.Parser)}
+	err := jwtVerifierURL.updateKeySet()
+
+	if err != nil {
+		log.Error().Err(err).Msg("unable to construct NewJWTVerifierURL")
```
mikecdavis: I think we can safely specify the message pertains to the "key set":
```go
log.Error().Err(err).Msg("unable to fetch key set")
```

PR #163, opened 2020-02-28
fix: Add documentation for auth & TLS functionality, plus related fixes

mjc1283:
## Summary

This provides documentation regarding authorization and TLS functionality recently added. This includes updates to the config properties table, an authorization guide under `/docs` , and a python example under `/examples`.

Also included are fixes for small issues uncovered while creating the documentation:
- Added support for multiple HMAC validation secrets, to facilitate zero-downtime rotation. The first secret in the array will be used to sign issued tokens
- Fix for an invalid response code (`0`) when the token issuer endpoint received invalid JSON in the request body

## JIRA
https://optimizely.atlassian.net/browse/OASIS-5899

### DIFF
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -1,0 +1,70 @@
+# Authorization Guide
+
+## Overview
+
+Optimizely Agent supports authorization workflows based on OAuth and JWT standards, allowing you to protect access to its API and Admin interfaces.
+
+There are three modes of operation:
+
+### Issuer & Validator
+Access tokens are issued by Agent itself, using a [Client Credentials grant](https://www.oauth.com/oauth2-servers/access-tokens/client-credentials/). Access tokens are signed and validated using the HS256 algorithm with a signing secret provided in configuration.
+
+### Validator-only
+Agent validates access tokens that were issued elsewhere. Access tokens are validated with public keys fetched from a [JWKS](https://tools.ietf.org/html/rfc7517) URL provided in configuration.
+
+### No authorization (default)
+The interface is publicly available.
+
+## Configuration
+- The API and Admin interfaces are each independently configured to run in one of the above-mentioned modes of operation.
+- Authorization configuration is applied under the `auth` key
+- Each mode of operation has its own set of configuration properties, described below.
+
+### Issuer & Validator
+The configuration properties pertaining to Issuer & Validator mode are listed below:
+
+|Property Name|Description|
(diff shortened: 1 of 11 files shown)


mjc1283:
@mikecdavis There is a Secret Rotation section at the bottom of the auth guide, but it's very short and I wasn't entirely sure what content we need there. What do you think?

docs/auth.md
```diff
@@ -0,0 +1,73 @@
...
+The configuration properties pertaining to Validator-only mode are listed below:
+
+|Property Name|Description|
+|---|---|
+|jwksURL|URL from which public keys should be fetched for token validation|
+
+### No authorization (default)
+The API & Admin interfaces run with no authorization when no `auth` configuration is given.
+
+### Example configuration file (yaml)
```
asaschachar: It seems like above there are two separate options, but here there is only one yaml file showing both examples. It might be more clear if we have two separate yaml examples for each option if I understand this correctly.

docs/auth.md
```diff
@@ -0,0 +1,73 @@
...
+## Overview
+
+Optimizely Agent supports authorization workflows based on OAuth and JWT standards, allowing you to protect access to its API and Admin interfaces.
+
+There are three modes of operation:
+
+### Issuer & Validator
+Access tokens are issued by Agent itself, using a [Client Credentials grant](https://www.oauth.com/oauth2-servers/access-tokens/client-credentials/). Access tokens are signed and validated using the HS256 algorithm with a signing secret provided in configuration.
+
+### Validator-only
```
asaschachar: If possible, it would be great to add one line to each option explaining why someone would choose one option over the other.

examples/auth.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/python
+# example: python auth.py <SDK-Key>
+
+import json
+import requests
+import sys
+
+# This example demonstrates obtaining an access token and using it to request the
```
asaschachar: which option does this correspond to? maybe be explicit in the comment?

docs/auth.md
mjc1283: Yep, that's right - I'll split it into two example yaml files.

examples/auth.py
mjc1283: Will fix it - this demonstrates Issuer & Validator mode.

PR #164, opened 2020-02-28
Matjaz/porting acceptance

Mat001:
@mikecdavis @juancarlostong 

It's not pretty. Lots to be cleaned out.  But it works locally.
Just cd tests/acceptance/ and run
pytest -v src/test_acceptance/ --host http://localhost:8080

Notes:
1. TWO .gitignore files now!
2. Directory structure of the tests may change - "src" may go out, need to figure out how to keep PYTHONPATH in place
3. @juancarlostong Can you pls check .travis.yaml change. Also I'd like this new stage to always pass for the . moment (not fail the build). Because I'm going to be changing things in the tests.


Mat001:
duplicate of 165

PR #168, opened 2020-03-05
fix: Use client secret hash in configuration for auth issuer & validator mode

mjc1283:
## Summary
- Change configuration for token issuance: client secret hash is provided via configuration instead of raw client secret. The expectation is that the hash in configuration is the base64-encoded BCrypt hash of the secret.
- Add helper tool to generate random secrets and associated hash
- Update auth documentation & example based on the above changes, and recommend against storing signing secret in config file

## Issues
https://optimizely.atlassian.net/browse/OASIS-5906


examples/auth.py
```diff
@@ -14,31 +14,33 @@
...
 # For more information, see docs/auth.md
 
 if len(sys.argv) < 2:
     sys.exit('Requires one argument: <SDK-Key>')
 
 sdk_key = sys.argv[1]
-client_id = sys.argv[2]
-client_secret = sys.argv[3]
+client_id = "clientid1"
+client_secret = "bvz7VWyxgLUySOZQpzSblsy3/8570JjIPKWO06SjliA="
```
mikecdavis: Did you mean to hardcode these? I'd recommend keeping them as runtime parameters, but adding the secret to an example above like with the `hmacSecrets`

pkg/jwtauth/jwtauth.go
```diff
@@ -57,11 +59,45 @@ func BuildAdminAccessToken(ttl time.Duration, key []byte) (tokenString string, e
...
+
+var bcryptWorkFactor = 12
+
+// ValidateClientSecret returns true if the hash of the secret provided in config matches
+// the secret provided in the request. Returns an error if the req secret fails base64
+// decoding.
+func ValidateClientSecret(reqSecret string, configSecretHash []byte) (bool, error) {
+	secretBytes, err := base64.StdEncoding.DecodeString(reqSecret)
+	if err != nil {
+		return false, fmt.Errorf("error decoding string: %v", err)
```
mikecdavis: Should this be `%w`?

pkg/handlers/oauth.go
```diff
@@ -67,9 +68,14 @@ func NewOAuthHandler(authConfig *config.ServiceAuthConfig) *OAuthHandler {
...
 	for _, clientCreds := range authConfig.Clients {
+		secretHashBytes, err := jwtauth.DecodeSecretHashFromConfig(clientCreds.SecretHash)
+		if err != nil {
+			log.Error().Err(err).Msgf("error decoding client creds secret (paired with client ID: %v)", clientCreds.ID)
+			continue
+		}
 		clientCredentials[clientCreds.ID] = ClientCredentials{
 			ID:     clientCreds.ID,
-			Secret: []byte(clientCreds.Secret),
+			Secret: secretHashBytes,
```
mikecdavis: Should we update to `SecretHash` or `SecretHashBytes`?

pkg/handlers/oauth.go
```diff
@@ -129,12 +135,24 @@ func (h *OAuthHandler) verifyClientCredentials(r *http.Request) (*ClientCredenti
...
+	if !ok {
+		return nil, http.StatusUnauthorized, &ClientCredentialsError{
+			ErrorCode:        "invalid_client",
+			ErrorDescription: "invalid client_id or client_secret",
+		}
+	}
+
+	isValid, err := jwtauth.ValidateClientSecret(reqBody.ClientSecret, clientCreds.Secret)
+	if err != nil {
+		middleware.GetLogger(r).Debug().Err(err).Msg("validating request secret")
```
mikecdavis: Should this be `Info`?

pkg/jwtauth/jwtauth.go
```diff
@@ -57,11 +59,45 @@ func BuildAdminAccessToken(ttl time.Duration, key []byte) (tokenString string, e
...
+	_, err = rand.Read(secretBytes)
+	if err != nil {
+		return "", "", fmt.Errorf("error returned from rand.Read: %v", err)
 	}
-	return subtle.ConstantTimeCompare(reqSecret, configSecret) == 1
+	secretStr = base64.StdEncoding.EncodeToString(secretBytes)
+
+	hashBytes, err := bcrypt.GenerateFromPassword(secretBytes, bcryptWorkFactor)
+	if err != nil {
+		return "", "", fmt.Errorf("error returned from bcrypt.GenerateFromPassword: %v", err)
```
mikecdavis: `%w`?

pkg/handlers/oauth.go
```diff
@@ -67,9 +68,14 @@ func NewOAuthHandler(authConfig *config.ServiceAuthConfig) *OAuthHandler {
 	clientCredentials := make(map[string]ClientCredentials)
 	// TODO: need to validate all client IDs are unique
 	for _, clientCreds := range authConfig.Clients {
+		secretHashBytes, err := jwtauth.DecodeSecretHashFromConfig(clientCreds.SecretHash)
+		if err != nil {
+			log.Error().Err(err).Msgf("error decoding client creds secret (paired with client ID: %v)", clientCreds.ID)
```
mikecdavis: Curious if this will fail if the customer is not using our generation tool.

pkg/jwtauth/jwtauth.go
```diff
@@ -57,11 +59,45 @@ func BuildAdminAccessToken(ttl time.Duration, key []byte) (tokenString string, e
...
+// Agent's authN/authZ workflow.
+// - The first return value is the secret - 32 random bytes, base64-encoded.
+// - The second return value is the bcrypt hash of the secret.
+// - The hash should be included in Agent's auth configuration as the client_secret value.
+// - The secret should be sent in the request to the token issuer endpoint.
+func GenerateClientSecretAndHash() (secretStr, hashStr string, err error) {
+	secretBytes := make([]byte, secretBytesLen)
+	_, err = rand.Read(secretBytes)
+	if err != nil {
+		return "", "", fmt.Errorf("error returned from rand.Read: %v", err)
```
mikecdavis: `%w`?

PR #178, opened 2020-03-10
/activate endpoint leniency

jmarsh410:
## Summary
This PR changes how the /activate endpoint handles experiment and feature keys that can't be found in the datafile. Instead of returning a 'key not found' message, it will return info detailing what specific key wasn't able to be found, while also returning info about the experiments that _were_ found.

**They "why"**
The /activate endpoint currently allows multiple experiment and feature keys to be passed up as params. However, if _any_ of those keys cannot be found in the current datafile an error is returned, even if some of the keys were able to be found.

It is very common that a code base includes code for some experiments that are turned on, and some that aren't, but if an app that uses this service wants to batch its calls together for performance and efficiency, then this makes it impossible to do so.

**An example**
A client-side app loads a page with any number of experiments running on it. It has a list of these existing experiments and uses that to make a request to /activate with the keys for all the features and experiments and awaits a response so that it can show the correct experience to the user. One of the experiments has not been turned on yet, or perhaps has just been turned off because it was complete or caused an unexpected error. Because of that the call to /activate will return an error and none of the experiments will work. The only alternative in this case is to make a separate call for each individual experiment, which takes longer and is less performant

### DIFF
--- a/api/openapi-spec/openapi.yaml
+++ b/api/openapi-spec/openapi.yaml
@@ -261,11 +261,7 @@ components:
           type: object
           additionalProperties:
             type: string
-        invalid:
-          type: boolean
-        key:
-          type: string
-        message:
+        error:
           type: string
       required:
         #        - id
--- a/api/openapi-spec/openapi.yaml
+++ b/api/openapi-spec/openapi.yaml
@@ -261,6 +261,12 @@ components:
           type: object
           additionalProperties:
             type: string
+        invalid:
+          type: boolean
+        key:
+          type: string
+        message:
+          type: string
       required:
         #        - id
         - key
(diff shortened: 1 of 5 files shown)


mikecdavis:
Thanks for the contribution! Please review the contributing [doc](https://github.com/optimizely/agent/blob/master/CONTRIBUTING.md) and sign our [Contributor License Agreement (CLA)](https://docs.google.com/a/optimizely.com/forms/d/e/1FAIpQLSf9cbouWptIpMgukAKZZOIAhafvjFCV8hS00XJLWQnWDFtwtA/viewform) while we review.

jmarsh410:
@mikecdavis Great idea, I've edited the Decision response to only add the error field

jmarsh410:
Thanks!

PR #179, opened 2020-03-10
Improve error messaging in override handler.

mikecdavis:
## Summary
- Checks OptimizelyConfig prior to setting overrides and error accordingly.
- Add more coverage to override handler and fix panic when payload is invalid
- Add additional helper method to `optimizelytest` pkg

### DIFF
--- a/pkg/optimizely/client.go
+++ b/pkg/optimizely/client.go
@@ -182,10 +182,21 @@ func (c *OptlyClient) SetForcedVariation(experimentKey, userID, variationKey str
 	if c.ForcedVariations == nil {
 		return false, ErrForcedVariationsUninitialized
 	}
+
+	// Check the entities exist as part of the Optimizely configuration
+	if optimizelyConfig := c.GetOptimizelyConfig(); optimizelyConfig == nil {
+		return false, errors.New("optimizely config is null")
+	} else if experiment, ok := optimizelyConfig.ExperimentsMap[experimentKey]; !ok {
+		return false, fmt.Errorf("experimentKey: %q %w", experimentKey, ErrEntityNotFound)
+	} else if _, ok := experiment.VariationsMap[variationKey]; !ok {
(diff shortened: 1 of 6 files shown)


pkg/optimizely/client.go
```diff
@@ -182,10 +182,21 @@ func (c *OptlyClient) SetForcedVariation(experimentKey, userID, variationKey str
...
+
+	// Check the entities exist as part of the Optimizely configuration
+	if optimizelyConfig := c.GetOptimizelyConfig(); optimizelyConfig == nil {
+		return false, errors.New("optimizely config is null")
+	} else if experiment, ok := optimizelyConfig.ExperimentsMap[experimentKey]; !ok {
+		return false, fmt.Errorf("experimentKey: %q %w", experimentKey, ErrEntityNotFound)
+	} else if _, ok := experiment.VariationsMap[variationKey]; !ok {
+		return false, fmt.Errorf("variationKey: %q %w", variationKey, ErrEntityNotFound)
+	}
+
```
mjc1283: We discussed unknown experiment/variation keys when implementing overrides in Go SDK, and ended up deciding to allow them to be set. IMO It can be useful to set an override for an experiment that will exist later after a datafile update.

But if we do this, probably we should fix it in Go SDK.

mjc1283: Or, maybe we can return something extra in the response to warn them that they set an invalid experiment/variation key.

mikecdavis: Yeah I almost made this PR just update the API spec to allow this, but I felt that the lack of feedback would be problematic. I like the suggestion to return some kind of message as opposed to an empty response and purely relying on status codes. So I'll make that change to "accept" non-existent experiment and variations, but will include something to that effect in the response.

mikecdavis: We should probably do something similar to the /track endpoint.

mikecdavis:
Closing in favor of https://github.com/optimizely/agent/pull/182

PR #180, opened 2020-03-10
fix: In API access token issuer, use SDK keys from configuration, not request header

mjc1283:
## Summary
`X-Optimziely-Sdk-Key` header is no longer required by token issuer endpoint. Instead, SDK keys are given in configuration of clients. 1 or more SDK keys are required for each client.

## JIRA
https://optimizely.atlassian.net/browse/OASIS-6073

### DIFF
--- a/pkg/handlers/oauth.go
+++ b/pkg/handlers/oauth.go
@@ -89,8 +89,6 @@ func NewOAuthHandler(authConfig *config.ServiceAuthConfig) *OAuthHandler {
 	}
 
 	h := &OAuthHandler{
-		// TODO: Interpret hmacSecret as a base64-format string, since we are already doing that for
-		// client secrets
 		hmacSecret:        []byte(hmacSecret),
 		ClientCredentials: clientCredentials,
 	}
--- a/pkg/handlers/oauth.go
+++ b/pkg/handlers/oauth.go
@@ -68,12 +68,11 @@ func NewOAuthHandler(authConfig *config.ServiceAuthConfig) *OAuthHandler {
 		if err != nil {
 			log.Error().Err(err).Msgf("error decoding client creds secret (paired with client ID: %v), skipping these credentials", clientCreds.ID)
 			continue
-
 		}
+
 		if len(clientCreds.SDKKeys) == 0 {
 			log.Error().Err(err).Msgf("client creds missing or empty SDK keys (with client ID: %v), skipping these credentials", clientCreds.ID)
 			continue
-
 		}
 
 		clientCredentials[clientCreds.ID] = ClientCredentials{
--- a/pkg/handlers/oauth.go
+++ b/pkg/handlers/oauth.go
@@ -89,6 +89,8 @@ func NewOAuthHandler(authConfig *config.ServiceAuthConfig) *OAuthHandler {
 	}
 
 	h := &OAuthHandler{
+		// TODO: Interpret hmacSecret as a base64-format string, since we are already doing that for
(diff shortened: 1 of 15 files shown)


pkg/handlers/oauth.go
```diff
@@ -70,13 +66,20 @@ func NewOAuthHandler(authConfig *config.ServiceAuthConfig) *OAuthHandler {
 	for _, clientCreds := range authConfig.Clients {
 		secretHashBytes, err := jwtauth.DecodeSecretHashFromConfig(clientCreds.SecretHash)
 		if err != nil {
-			log.Error().Err(err).Msgf("error decoding client creds secret (paired with client ID: %v)", clientCreds.ID)
+			log.Error().Err(err).Msgf("error decoding client creds secret (paired with client ID: %v), skipping these credentials", clientCreds.ID)
+			continue
+		}
+
+		if len(clientCreds.SDKKeys) == 0 {
+			log.Error().Err(err).Msgf("client creds missing or empty SDK keys (with client ID: %v), skipping these credentials", clientCreds.ID)
```
mjc1283: Note: getting an access token for "Any SDK key" is not supported in this implementation. A specific set of SDK keys must be set, or the client will be skipped.

mikecdavis: Are there plans for global support?

pkg/middleware/auth.go
```diff
@@ -271,9 +271,31 @@ func (a Auth) AuthorizeAPI(next http.Handler) http.Handler {
 				RenderError(errors.New("token expired"), http.StatusUnauthorized, w, r)
 				return
 			}
+
+			rawClaimsSdkKeys, ok := claims["sdk_keys"].([]interface{})
+			if !ok {
+				RenderError(errors.New("Invalid claims: sdk_keys not found, or have the wrong type"), http.StatusUnauthorized, w, r)
+				return
+			}
+			claimsSdkKeys := make([]string, len(rawClaimsSdkKeys))
```
mikecdavis: I don't think this slice is necessary since you can validate the requisite param while iterating over the `rawClaimsSdkKeys`. So a single loop is all you need.

pawels-optimizely: agree

pkg/handlers/oauth.go
mjc1283: No. My intuition is that we shouldn't support it, but I am open to being persuaded otherwise.

mikecdavis: That's fine for now, we can add this option in the future.

pkg/middleware/auth.go
```diff
@@ -271,9 +271,27 @@ func (a Auth) AuthorizeAPI(next http.Handler) http.Handler {
...
-				RenderError(errors.New("SDK keys not equal"), http.StatusUnauthorized, w, r)
+			foundMatchingSdkKey := false
+			for _, rawSdkKey := range rawClaimsSdkKeys {
+				sdkKey, ok := rawSdkKey.(string)
+				if !ok {
+					GetLogger(r).Warn().Msgf("Non-string value in token claims sdk_keys: %v", sdkKey)
+					continue
+				}
+				if sdkKey == sdkKeyFromHeader {
+					foundMatchingSdkKey = true
```
mikecdavis: Nit, I think this can be refactored so you don't need `foundMatchingSdkKey`. I've been looking for a chance to use [labels](https://medium.com/golangspec/labels-in-go-4ffd81932339) :)

PR #185, opened 2020-03-12
feature (examples): add notification example

thomaszurkan-optimizely:
## Summary
- Fix basic to call isFeatureEnabled instead of activate on feature key
- Add a notification example

### DIFF
--- a/examples/basic.py
+++ b/examples/basic.py
@@ -27,7 +27,7 @@ payload = {
 
 for key in env['featuresMap']:
     params = {"featureKey": key}
-    resp = s.post(url = 'http://localhost:8080/v1/activate', params=params, json=payload)
+    resp = s.post(url = 'http://localhost:8080/v1/isFeatureEnabled', params=params, json=payload)
     print("Feature Key: {}".format(key))
     print(json.dumps(resp.json()[0], indent=4, sort_keys=True))
 
--- a/examples/basic.py
+++ b/examples/basic.py
@@ -27,7 +27,7 @@ payload = {
 
 for key in env['featuresMap']:
     params = {"featureKey": key}
-    resp = s.post(url = 'http://localhost:8080/v1/isFeatureEnabled', params=params, json=payload)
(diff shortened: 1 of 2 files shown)


examples/basic.py
```diff
@@ -27,7 +27,7 @@
 
 for key in env['featuresMap']:
     params = {"featureKey": key}
-    resp = s.post(url = 'http://localhost:8080/v1/activate', params=params, json=payload)
+    resp = s.post(url = 'http://localhost:8080/v1/isFeatureEnabled', params=params, json=payload)
```
mikecdavis: This is not an error. /v1/isFeatureEnabled does not exist and everything is routed through `/activate` for both features and experiments.

examples/notifications.py
```diff
@@ -0,0 +1,21 @@
+from sseclient import SSEClient
+from httplib import IncompleteRead
+
+def get_messages():
+    requests_kwargs = {'headers': {}}
+    requests_kwargs['headers']['X-Optimizely-SDK-Key'] = 'FeEqVPw2ZKLcxjHX5A732L'
```
mikecdavis: Not used.

examples/notifications.py
```diff
@@ -0,0 +1,21 @@
...
+    print('here')
+    messages = SSEClient('http://localhost:8080/v1/notifications/event-stream', timeout=10000, chunk_size=128,
+                         headers={'X-Optimizely-SDK-Key': 'FeEqVPw2ZKLcxjHX5A732L'})
```
mikecdavis: Can you make this a parameter like the other examples in this directory?

examples/notifications.py
```diff
@@ -0,0 +1,21 @@
+from sseclient import SSEClient
+from httplib import IncompleteRead
```
mikecdavis: Please include a header and example usage like the other examples in this directory.

mikecdavis:
Thanks @thomaszurkan-optimizely for making these changes, I see you addressed my comments yesterday. That being said @mjc1283 just put up a PR with a similar example also with the documentation, so we can probably close this.

PR #187, opened 2020-03-13
fix(auth): HMAC secret strings from config are interpreted as base64

mjc1283:
## Summary

Introduce an expectation that `hmacSecrets` config values are in base64 format. The rationale is that we already have this expectation for client secret hashes in config, and we should be consistent with how we interpret and convert config strings into byte slices.

## Test Plan
Manual testing, updated unit tests, updated example and auth guide

### DIFF
--- a/pkg/handlers/oauth.go
+++ b/pkg/handlers/oauth.go
@@ -64,7 +64,7 @@ func NewOAuthHandler(authConfig *config.ServiceAuthConfig) *OAuthHandler {
 	clientCredentials := make(map[string]ClientCredentials)
 	// TODO: need to validate all client IDs are unique
 	for _, clientCreds := range authConfig.Clients {
-		secretHashBytes, err := jwtauth.DecodeSecretHashFromConfig(clientCreds.SecretHash)
+		secretHashBytes, err := jwtauth.DecodeConfigValue(clientCreds.SecretHash)
 		if err != nil {
 			log.Error().Err(err).Msgf("error decoding client creds secret (paired with client ID: %v), skipping these credentials", clientCreds.ID)
 			continue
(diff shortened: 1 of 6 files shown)


pkg/handlers/oauth.go
```diff
@@ -83,18 +83,26 @@ func NewOAuthHandler(authConfig *config.ServiceAuthConfig) *OAuthHandler {
...
-		hmacSecret = authConfig.HMACSecrets[0]
+	hmacSigningSecret := []byte{}
+	for i, hmacSecret := range authConfig.HMACSecrets {
+		decodedSecret, err := jwtauth.DecodeConfigValue(hmacSecret)
+		if err != nil {
+			log.Error().Err(err).Msgf("error decoding HMAC secret")
+			return nil
+		}
+		if i == 0 {
+			// The first secret is used to sign tokens - the rest are only used for validation (see middleware/auth.go)
```
pawels-optimizely: do we want to really exit when we have one secret that cannot be decoded and many more that can be decoded? 
Can we flag (delete) bad secrets and see what we have left and then assign hmacSigningSecret to the first good secret ?

pkg/middleware/auth.go
```diff
@@ -325,6 +326,16 @@ func NewAuth(authConfig *config.ServiceAuthConfig) *Auth {
 		return &Auth{Verifier: NoAuth{}}
 	}
 
-	return &Auth{Verifier: NewJWTVerifier(authConfig.HMACSecrets)}
+	decodedSecrets := [][]byte{}
+	for _, hmacSecret := range authConfig.HMACSecrets {
+		decodedSecret, err := jwtauth.DecodeConfigValue(hmacSecret)
+		if err != nil {
+			log.Error().Err(err).Msg("failed to decode HMAC secret")
```
pawels-optimizely: same comment as before. if we have one bad out of many good ones, do we still disable (return nil)?

pkg/handlers/oauth.go
mjc1283: I changed my mind since I implemented the "skipping invalid" pattern above for client secrets. I think the user wants to know ASAP if something is invalid. If we start and skip invalid with error log, it is easier for them to miss it and would be frustrating to find later.

mjc1283: If we are all in agreement, I should definitely change it to be the same above for client secrets.

mjc1283: Created follow-up task: https://optimizely.atlassian.net/browse/OASIS-6108

PR #189, opened 2020-03-17
fix(cache): Adds flushInterval to client loader

mikecdavis:
## Summary
- Fix omission of passing flushInterval config
- Refactor to facilitate testing

## Issues
- OASIS-6097

### DIFF
--- a/pkg/optimizely/cache.go
+++ b/pkg/optimizely/cache.go
@@ -109,9 +109,13 @@ func initOptlyClient(sdkKey string, conf config.ClientConfig, metricsRegistry *M
 	}
 
 	q := event.NewInMemoryQueue(conf.QueueSize)
-	ep := event.NewBatchEventProcessor(event.WithQueueSize(conf.QueueSize),
-		event.WithBatchSize(conf.BatchSize), event.WithQueue(q),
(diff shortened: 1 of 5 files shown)


pkg/optimizely/cache.go
```diff
@@ -34,23 +34,25 @@ import (
...
 	cache := &OptlyCache{
-		ctx:             ctx,
-		wg:              sync.WaitGroup{},
-		loader:          initOptlyClient,
-		optlyMap:        cmap.New(),
-		clientConf:      conf,
-		metricsRegistry: metricsRegistry,
+		ctx:      ctx,
+		wg:       sync.WaitGroup{},
+		loader:   defaultLoader(conf, metricsRegistry, cmLoader, event.NewBatchEventProcessor),
```
mjc1283: Re, the TODO comment, what happens if we do this?
```suggestion
		loader:   defaultLoader(conf, metricsRegistry, sdkconfig.NewPollingProjectConfigManager, event.NewBatchEventProcessor),
```

mikecdavis: That's what I want to do, but that throws a compiler error. The method above coerces into the interface.

PR #193, opened 2020-03-18
Serve openapi.yaml statically.

mikecdavis:
## Summary
- Embeds openapi.yaml into go executable
- Serves openapi.yaml from api router
- Adds basic index handler

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ clean: check-go ## runs `go clean` and removes the bin/ dir
 	$(GOCLEAN) --modcache
 	rm -rf $(GOBIN)
 
-cover: check-go ## runs test suite with coverage profiling
+cover: check-go static ## runs test suite with coverage profiling
 	$(GOTEST) ./... -coverprofile=$(COVER_FILE)
 
 cover-html: cover ## generates test coverage html report
@@ -53,7 +53,7 @@ install: check-go ## installs all dev and ci dependencies, but does not install 
 	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(GOPATH)/bin v1.19.0
 	go get github.com/rakyll/statik
 
-lint: check-go ## runs `golangci-lint` linters defined in `.golangci.yml` file
+lint: check-go static ## runs `golangci-lint` linters defined in `.golangci.yml` file
 	$(GOLINT) run --out-format=tab --tests=false ./...
 
 run: $(TARGET) ## builds and executes the TARGET binary
@@ -62,7 +62,7 @@ run: $(TARGET) ## builds and executes the TARGET binary
 static: check-go
 	statik -src=api/openapi-spec
 
-test: check-go ## recursively tests all .go files
+test: check-go static ## recursively tests all .go files
 	$(GOTEST) ./...
 
 include scripts/Makefile.ci
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ LDFLAGS=-ldflags "-s -w -X main.Version=${APP_VERSION} -X github.com/optimizely/
 
 all: test lint build ## runs the test, lint and build targets
 
-$(TARGET): check-go
+$(TARGET): check-go static
 	$(GOBUILD) $(LDFLAGS) -o $(GOBIN)/$(TARGET) cmd/optimizely/main.go
 
 build: $(TARGET) check-go ## builds and installs binary in bin/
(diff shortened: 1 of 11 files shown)


pkg/routers/api.go
```diff
@@ -130,4 +132,12 @@ func WithAPIRouter(opt *APIOptions, r chi.Router) {
...
 
 	r.With(createAccesstokenTimer).Post("/oauth/token", opt.oAuthHandler)
+
+	statikFS, err := fs.New()
+	if err != nil {
+		panic(err)
+	}
+
+	staticServer := http.FileServer(statikFS)
+	r.Handle("/openapi.yaml", staticServer)
```
tylerbrandt: Is this a conventional place to look for "what is the API spec of this service"? If not, is there such a place (something akin to https://en.wikipedia.org/wiki/List_of_/.well-known/_services_offered_by_webservers but for OpenAPI)? If not, would we be willing to serve it on "/" (index)? Seems a little circular that I have to "just know" this one route in order to discover the rest of the routes I can use.

mikecdavis: From https://swagger.io/specification/
> It is RECOMMENDED that the root OpenAPI document be named: openapi.json or openapi.yaml.

I also found other examples in their [documentation](https://swagger.io/docs/specification/api-host-and-base-path/#relative-urls) like:
> For example, if the definition hosted at http://localhost:3001/openapi.yaml...

mikecdavis: That being said I think it's a great idea to have an index "/" handler that could say "hello" and link to the spec. I'll add that as it was also mentioned by @asaschachar as a nice to have to improve the overall "landing page" experience.

Makefile
```diff
@@ -43,22 +43,26 @@ clean: check-go ## runs `go clean` and removes the bin/ dir
...
-cover: check-go ## runs test suite with coverage profiling
+cover: check-go static ## runs test suite with coverage profiling
 	$(GOTEST) ./... -coverprofile=$(COVER_FILE)
 
 cover-html: cover ## generates test coverage html report
 	$(GOCMD) tool cover -html=$(COVER_FILE)
 
 install: check-go ## installs all dev and ci dependencies, but does not install golang
 	curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(GOPATH)/bin v1.19.0
+	go get github.com/rakyll/statik
```
mjc1283: Why isn't this automatically installed like other deps?

mikecdavis: This is a tool to embed static files into the binary. This command is installing the `statik` CLI which is run under the `static` make target.

pawels-optimizely: why do we need this package?  can we server directly using staticServer := http.FileServer(http.Dir(filesDir)) ?

mikecdavis: So we can still deploy a static binary and not have to incorporate this content into each deployment. Excerpt from the statik repo:
> Is this a crazy idea? No, not necessarily. If you're building a tool that has a Web component, you typically want to serve some images, CSS and JavaScript. You like the comfort of distributing a single binary, so you don't want to mess with deploying them elsewhere. If your static files are not large in size and will be browsed by a few people, statik is a solution you are looking for.

web/static/index.html
```diff
@@ -0,0 +1,44 @@
...
+
+<html lang="en">
+<head>
+    <meta charset="utf-8">
+
+    <title>Optimizely Agent</title>
+    <meta name="description" content="Optimizely Agent">
+    <meta name="author" content="Optimizely">
+
+    <link rel="stylesheet" href="styles.css">
```
mjc1283: styles.css doesn't exist?

web/static/index.html
```diff
@@ -0,0 +1,44 @@
...
+
+</head>
+
+<body>
+<h1>Welcome to Optimizely Agent!</h1>
```
asaschachar: Yes! This is exactly what this project needed!

web/static/index.html
```diff
@@ -0,0 +1,44 @@
...
+    .#######^            '"#@@@@@##|#@@@####"`
+     #####   ,##m           ;|5#T"*@###5\
+     '#M   ;######M     ,;####`  ,#####
+         ;######"   ;######M   ;######`
+        \#####"   s######"   ;######"
+          "#^  ,#######"   s######"
+              %######`  ,#####M"
+                  ``   """`
+</pre>
+Please refer to the <a href="/openapi.yaml">openapi.yaml</a> spec for API details.
```
asaschachar: Cool, I think this is good to keep as one link. We can evaluate whether we need more links, but I actually think keeping this lean is important for now so we don't duplicate documentation efforts across different surfaces

PR #201, opened 2020-03-23
ci(travis): include statik dependency

juancarlostong:
## Summary
- due to new statik depedency, added extra line to run `make install` to the build Dockerfiles
- made uploading packages contingent on successful package creation (previously it was uploading regardless)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,10 +105,8 @@ jobs:
       env: APP_VERSION=${TRAVIS_TAG#v}
 
       script:
-        # now we're going to create packages
-        - $TRAVIS_BUILD_DIR/scripts/ci_create_packages.sh
-        # upload packages to artifactory
-        - $TRAVIS_BUILD_DIR/scripts/ci_upload_packages.sh
+        # now we're going to create packages & upload packages
+        - $TRAVIS_BUILD_DIR/scripts/ci_create_packages.sh && $TRAVIS_BUILD_DIR/scripts/ci_upload_packages.sh
 
 before_script:
   # https://github.com/travis-ci/gimme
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -105,10 +105,8 @@ jobs:
       env: APP_VERSION=${TRAVIS_TAG#v}
 
       script:
-        # now we're going to create packages
-        - $TRAVIS_BUILD_DIR/scripts/ci_create_packages.sh
-        # upload packages to artifactory
-        - $TRAVIS_BUILD_DIR/scripts/ci_upload_packages.sh
+        # now we're going to create packages & upload packages
+        - $TRAVIS_BUILD_DIR/scripts/ci_create_packages.sh && $TRAVIS_BUILD_DIR/scripts/ci_upload_packages.sh
```
juancarlostong: this ought to stop it from uploading if packages failed to be created

scripts/dockerfiles/Dockerfile.static
```diff
@@ -3,6 +3,7 @@ FROM golang:$GO_VERSION as builder
 
 WORKDIR /go/src/github.com/optimizely/agent
 COPY . .
+RUN make install
```
juancarlostong: this drops in the new dependency on statik

.travis.yml
```diff
@@ -93,7 +93,16 @@ jobs:
...
+      name: linux
+      os: linux
+      dist: xenial
+
+      script:
+        - export APP_VERSION=$(git describe --abbrev=0 --tags | tr -d '^v')
+        - $TRAVIS_BUILD_DIR/scripts/ci_create_packages.sh
+
+    - stage: Build and Upload
       if: type = push AND tag IS present
```
juancarlostong: added a new stage to test building docker images without uploading them... less clunky debugging

msohailhussain: I will recommend to use `wait-for-it` to check port is opened.

juancarlostong: agent is not being run, its just built and packaged so there's no port involved here

PR #219, opened 2020-04-23
feat(datafile): Support for DatafileURLTemplate.

yasirfolio3:
## Summary
- Added support to pass through DatafileURLTemplate from the Agent config to the Go SDK project config manager.

## Test Plan
- FSC should pass.

## Issues
- https://optimizely.atlassian.net/browse/OASIS-6329

### DIFF
--- a/cmd/optimizely/main_test.go
+++ b/cmd/optimizely/main_test.go
@@ -47,6 +47,7 @@ func assertClient(t *testing.T, actual config.ClientConfig) {
 	assert.Equal(t, 1, actual.BatchSize)
 	assert.Equal(t, 10, actual.QueueSize)
 	assert.Equal(t, 1*time.Minute, actual.FlushInterval)
+	assert.Equal(t, "https://localhost/v1/%s.json", actual.DatafileURLTemplate)
 }
 
 func assertLog(t *testing.T, actual config.LogConfig) {
@@ -141,6 +142,7 @@ func TestViperProps(t *testing.T) {
 	v.Set("client.batchSize", 1)
 	v.Set("client.queueSize", 10)
 	v.Set("client.flushInterval", 1*time.Minute)
+	v.Set("client.datafileURLTemplate", "https://localhost/v1/%s.json")
 
 	v.Set("log.pretty", true)
 	v.Set("log.level", "debug")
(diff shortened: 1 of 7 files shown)


config.yaml
```diff
@@ -91,3 +91,5 @@ client:
     queueSize: 1000
     ## the maximum time between events being dispatched
     flushInterval: 30s
+    ## the endpoint for retrieving the datafile from the CDN
+    datafileURLTemplate: "https://cdn.optimizely.com/datafiles/%s.json"
```
mikeng13, line 95: This is defined in the Go SDK publicly here: https://github.com/optimizely/go-sdk/blob/master/pkg/config/polling_manager.go#L4 please reference that instead

msohailhussain: eol

config.yaml
```diff
@@ -91,3 +91,5 @@ client:
     queueSize: 1000
     ## the maximum time between events being dispatched
     flushInterval: 30s
+    ## the endpoint for retrieving the datafile from the CDN
```
mikecdavis, line 94: Let's use this to describe the formatting.
> Template URL for SDK datafile location. The template should specify a "%s" token for SDK key substitution.

cmd/optimizely/main_test.go
```diff
@@ -215,6 +217,7 @@ func TestViperEnv(t *testing.T) {
 	_ = os.Setenv("OPTIMIZELY_CLIENT_BATCHSIZE", "1")
 	_ = os.Setenv("OPTIMIZELY_CLIENT_QUEUESIZE", "10")
 	_ = os.Setenv("OPTIMIZELY_CLIENT_FLUSHINTERVAL", "1m")
+	_ = os.Setenv("OPTIMIZELY_CLIENT_DATAFILEURLTEMPLATE", "https://localhost/v1/%s.json")
```
msohailhussain, line 220: @mikecdavis 
Shouldn't it be `DATAFILE_URL_TEMPLATE`

mikecdavis: No this needs to be `DATAFILEURLTEMPLATE` based on how viper maps config values to ENV variables.

PR #221, opened 2020-04-30
feat(cors): Added cross origin support

msohailhussain:
## Summary
- Cross origin support added specifically when running from js.

## TestPlan
- Unit tests added that checks headers for preflight requests.

## Issues
No Issues


pkg/routers/api_test.go
```diff
@@ -142,6 +166,131 @@ func (suite *APIV1TestSuite) TestCreateAccessToken() {
...
+		suite.Equal(validOrigin, rec.Header().Get(corsOriginHeaderKey))
+	}
+
+	// Unallowed Origin
+	for _, route := range routes {
+		req := httptest.NewRequest(route.method, "/v1/"+route.path, nil)
+		req.Header.Add(originHeaderKey, "http://test.com")
+		rec := httptest.NewRecorder()
+		suite.mux.ServeHTTP(rec, req)
+		suite.Equal(http.StatusOK, rec.Code)
```
msohailhussain, line 197: Why it's OK? it shouldn't be OK.

config/config.go
```diff
@@ -121,12 +134,35 @@ type ServerConfig struct {
...
+	AllowedOrigins     []string `json:"allowedOrigins"`
+	AllowedMethods     []string `json:"allowedMethods"`
+	AllowedHeaders     []string `json:"allowedHeaders"`
+	ExposedHeaders     []string `json:"exposedHeaders"`
+	AllowedCredentials bool     `json:"allowedCredentials"`
+	MaxAge             int      `json:"maxAge"`
+}
+
+// ToCorsOptions converts CORSConfig to cors middleware options
+func (c *CORSConfig) ToCorsOptions() cors.Options {
```
mikecdavis, line 155: Nit, but I'd prefer we didn't add this here as it feels like we're leaking a dependency. I think having a method in `api.go` would be a cleaner abstraction.
```
func CreateCorsHandler(c conf.CORSConfig) func(next http.Handler) http.Handler {
  options := cors.Options{
		AllowedOrigins:   c.AllowedOrigins,
		AllowedMethods:   c.AllowedMethods,
		AllowedHeaders:   c.AllowedHeaders,
		ExposedHeaders:   c.ExposedHeaders,
		AllowCredentials: c.AllowedCredentials,
		MaxAge:           c.MaxAge,
	}

  return cors.Handler(options)
}
```

config/config.go
```diff
@@ -47,6 +49,17 @@ func NewDefaultConfig() *AgentConfig {
 				JwksURL:            "",
 				JwksUpdateInterval: 0,
 			},
+			CORS: CORSConfig{
+				// If allowedOrigins is nil or empty, value is set to ["*"].
+				AllowedOrigins: nil,
```
mikecdavis, line 54: Looks like this is defaulting to wide-open CORS support. We need to default to disabled then enable only if explicitly configured.

pawels-optimizely: agree with this comment

mjc1283:
@msohailhussain @mikecdavis I suggest this setting be used in the Notifications handler, which is [currently hard-coded to `Access-Control-Allow-Origin: *`](https://github.com/optimizely/agent/blob/0f1e0028a91d9ff85ae906d35a670606d69a224d/pkg/handlers/notification.go#L84).

config/config.go
msohailhussain: Discussed offline, it's OK to comment it out and give an idea to the user, CORS is not enabled. It will work as expected without `cors` before implementing it how it was working.

config.yaml
```diff
@@ -51,14 +51,32 @@ api:
...
+    #     allowedCredentials: false
+    #     maxAge: 300
 ##
 ## admin service configuration
 ##
 admin:
     ## http listener port
     port: "8088"
-
+
```
msohailhussain, line 79: remove this line.

msohailhussain:
https://github.com/optimizely/agent/pull/221#issuecomment-623566894
I will take care of it @mjc1283

pkg/routers/api_test.go
```diff
@@ -66,6 +78,17 @@ var testAuthMiddleware = func(next http.Handler) http.Handler {
 	return http.HandlerFunc(fn)
 }
 
+var testCorsHandler = cors.Handler(
```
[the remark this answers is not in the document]
msohailhussain: @mikecdavis handler is added for all requests. The only way to verify is to check things are coming back in the response or not. 
Please suggest if you want unit tests other than this approach.

msohailhussain:
@mjc1283 I added this commit, https://github.com/optimizely/agent/pull/221/commits/cca13092d310ceda2f31adcc164485c55eedd558
AccessControl header is set in parent handler. And it's only useful for OPTION requests, not for all.

config/config.go
```diff
@@ -47,6 +47,17 @@ func NewDefaultConfig() *AgentConfig {
 				JwksURL:            "",
 				JwksUpdateInterval: 0,
 			},
+			CORS: CORSConfig{
+				// If AllowedOrigins is nil or empty, value is set to ["*"].
+				AllowedOrigins: nil,
+				// If AllowedMethods is nil or empty, value is set to (HEAD, GET and POST).
+				AllowedMethods: []string{"HEAD", "GET", "PUT", "POST", "DELETE"},
```
mikecdavis, line 54: Might as well default to `nil` or empty here rather than to _extend_ the default.

msohailhussain: Sure, will do.

PR #226, opened 2020-05-15
refact(feature-test): Returning experimentKey and variationKey with feature test decisions.

yasirfolio3:
## Summary
- Returning experimentKey and variationKey with feature test decisions.

## Test Plan
- FSC should pass.

## Issues
https://optimizely.atlassian.net/browse/OASIS-6247


pkg/optimizely/client.go
```diff
@@ -175,6 +175,22 @@ func (c *OptlyClient) ActivateFeature(key string, uc entities.UserContext, disab
 		return &Decision{}, err
 	}
 
+	var experimentKey, variationKey string
+	projectConfig, _ := c.OptimizelyClient.ConfigManager.GetConfig()
+
+	feature, _ := projectConfig.GetFeatureByKey(key)
```
msohailhussain, line 181: As per comment from PR # 224, please use OptimizelyConfig.

yasirfolio3: Since we already need `projectConfig` to create `decisionContext`, we are just reusing it here for `feature`.

pkg/optimizely/client.go
```diff
@@ -175,22 +175,17 @@ func (c *OptlyClient) ActivateFeature(key string, uc entities.UserContext, disab
...
-	// HACK - Triggers impression events when applicable. This is not
-	// ideal since we're making TWO decisions for each feature now. TODO OASIS-5549
-	if !disableTracking {
-		_, tErr := c.IsFeatureEnabled(key, uc)
-		if tErr != nil {
-			return &Decision{}, tErr
-		}
-	}
+	// This is not ideal since we're making TWO decisions for each feature now. TODO OASIS-5549
+	experimentKey, variationKey := c.GetFeatureDecisionKeysAndTrack(key, uc, disableTracking)
```
msohailhussain, line 179: `GetDetailedFeatureDecisionUnsafe`

pkg/handlers/activate_test.go
```diff
@@ -125,10 +127,12 @@ func (suite *ActivateTestSuite) TestTrackFeatureWithFeatureRollout() {
...
-		UserID:     "testUser",
-		FeatureKey: "one",
-		Enabled:    true,
-		Type:       "feature",
+		UserID:        "testUser",
+		FeatureKey:    "one",
+		Enabled:       true,
+		Type:          "feature",
+		ExperimentKey: "1",
+		VariationKey:  "3",
```
mjc1283, line 135: As mentioned in the Go SDK PR: I don't think it should return `ExperimentKey` and `VariationKey` when the user was bucketed into a rollout (not a feature test). The keys of rollout experiments are not exposed or meaningful to the user.

go.mod
```diff
@@ -11,7 +11,8 @@ require (
 	github.com/go-kit/kit v0.9.0
 	github.com/google/uuid v1.1.1
 	github.com/lestrrat-go/jwx v0.9.0
-	github.com/optimizely/go-sdk v1.2.0
+	// Using custom go-sdk branch for testing purposes.
+	github.com/optimizely/go-sdk v1.1.3-0.20200604162003-16569aea2cdc
```
mjc1283, line 15: I think we should first create a Go SDK release that includes the new method, then update the dependency here to the new Go SDK version.

PR #229, opened 2020-05-21
chore: Update Go SDK to v1.2.0, prepare changelog for release

mjc1283:
## Summary
- Update Go SDK dependency to v1.2.0 (add support for multi-rule rollouts)
- Revert usage of method not available in Go SDK v.1.2.0
- Update CHANGELOG to prepare for v1.1.0 Agent release

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -95,6 +95,8 @@ github.com/optimizely/go-sdk v1.1.3-0.20200504184941-ac759bfb12f8 h1:EO1/LbxE2C4
 github.com/optimizely/go-sdk v1.1.3-0.20200504184941-ac759bfb12f8/go.mod h1:ehZPiEzNzNJw98EFrZX1C9h9gVgVRCiWosQEVlgivCw=
 github.com/optimizely/go-sdk v1.1.3 h1:zvhrj+nNlCrndXe7DbFtTQiy9Rny6ZLMyqdap1TMFi8=
 github.com/optimizely/go-sdk v1.1.3/go.mod h1:aA/UeFjLeQefRlvfTI8QkvBmJWPvLEHamKmp/CdJqGU=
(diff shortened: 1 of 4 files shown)


pkg/optimizely/client.go
```diff
@@ -170,7 +170,7 @@ func (c *OptlyClient) RemoveForcedVariation(experimentKey, userID string) (*Over
 // ActivateFeature activates a feature for a given user by getting the feature enabled status and all
 // associated variables
 func (c *OptlyClient) ActivateFeature(key string, uc entities.UserContext, disableTracking bool) (*Decision, error) {
-	enabled, variables, err := c.GetAllFeatureVariablesWithDecision(key, uc)
+	enabled, variables, err := c.GetAllFeatureVariables(key, uc)
```
aliabbasrizvi, line 173: Was this anticipated?

go.sum
```diff
@@ -95,6 +95,8 @@ github.com/optimizely/go-sdk v1.1.3-0.20200504184941-ac759bfb12f8 h1:EO1/LbxE2C4
 github.com/optimizely/go-sdk v1.1.3-0.20200504184941-ac759bfb12f8/go.mod h1:ehZPiEzNzNJw98EFrZX1C9h9gVgVRCiWosQEVlgivCw=
 github.com/optimizely/go-sdk v1.1.3 h1:zvhrj+nNlCrndXe7DbFtTQiy9Rny6ZLMyqdap1TMFi8=
 github.com/optimizely/go-sdk v1.1.3/go.mod h1:aA/UeFjLeQefRlvfTI8QkvBmJWPvLEHamKmp/CdJqGU=
```
mikecdavis, line 97: Can you run `go mod tidy` to clean this up?

pkg/optimizely/client.go
mjc1283: Yes, this is removing a method that was only available in the previous dependency version of Go SDK, and not in v1.2.0.

PR #232, opened 2020-06-02
feat(health): Exposing health endpoint on the public api side

yasirfolio3:
## Summary
- Made health endpoint public.

## Test Plan
- FSC should pass.

## Issues
- https://optimizely.atlassian.net/browse/OASIS-6454


pkg/handlers/admin_entities.go
```diff
@@ -33,8 +33,8 @@ var startTime = time.Now()
 // JSON is a map alias, just for convenience
 type JSON map[string]interface{}
 
-// Health is holding info about health checks
-type Health struct {
+// HealthBody is holding info about health checks
+type HealthBody struct {
```
msohailhussain, line 37: Why changed name?

pkg/handlers/admin_entities_test.go
```diff
@@ -34,7 +34,7 @@ var testConfig = config.AgentConfig{
 	Name:    "3",
 }
 
-func TestHealthHandlerBothServicesStarted(t *testing.T) {
+func TestAdminHealthHandlerBothServicesStarted(t *testing.T) {
```
msohailhussain, line 37: Why renamed it?

pkg/handlers/admin_entities.go
yasirfolio3: since we have declared a new handler called `Health` here, we cannot have a struct with the same name: https://github.com/optimizely/agent/pull/232/files#diff-dd19b841dc70477024a581f7481b9784R27

pkg/handlers/admin_entities_test.go
yasirfolio3: since a new test with a similar name was added here: https://github.com/optimizely/agent/pull/232/files#diff-06adbb118deb5fc1de8da69c3b3668b6R29

pkg/handlers/admin_entities.go
aliabbasrizvi: I'd prefer something like `HealthData` or `HealthInfo` over `HealthBody`.

mikecdavis: Let's move this struct into the new `Health` handler and remove this `admin.Health` handler. That would mean modifying `routers/admin.go` to use this new `Health` handler.

pkg/routers/api.go
```diff
@@ -135,6 +138,7 @@ func WithAPIRouter(opt *APIOptions, r chi.Router) {
 		r.With(opt.oAuthMiddleware).Get("/notifications/event-stream", opt.nStreamHandler)
 	})
 
+	r.With(healthTimer).Get("/health", opt.healthHandler)
```
mikecdavis, line 141: We also need to make this endpoint configurable from within `config.go`. This applies to the api listener as well as the admin listener.

PR #234, opened 2020-06-15
Streamline CI

mikecdavis:
## Summary
- Remove awesome-bot
- Remove mdspell
- Remove RPM and DEB builds
- Remove CA Cert pinning in alpine build
- Force etcd indirect dependency to the latest patch version

These markdown linters are adding more friction than the problem they aim to resolve so we removed them. We ran into build failures for RPM packages due to an insufficient ruby version so we opted to remove RPM and DEB builds until there is a demand. SourceClear started to flag an indirect dependency as a vulnerability, but unfortunately etcd @ v3.4.x is broken in go mod so the best we could do was bump to the latest 3.3.x version. Lastly, we needed to remove the cert pinning in the alpine docker build since that version was no longer available.

Ref: https://github.com/etcd-io/etcd/issues/11154


go.mod
```diff
@@ -30,5 +30,11 @@ exclude (
 	github.com/coreos/etcd v3.3.12+incompatible
 	github.com/coreos/etcd v3.3.13+incompatible
 	github.com/coreos/etcd v3.3.15+incompatible
+	github.com/coreos/etcd v3.3.16+incompatible
```
mikeng13, line 33: Lol this seems a bit excessive

mikecdavis: Haha, yeah. There is a `replace` directive in go mod. And while it works to force 3.3.10 => 3.3.22 in a single line, srcclr doesn't appear to respect that and still flags the original version 3.3.10 which has even higher severity warnings.

PR #236, opened 2020-06-19
feat: run docker container as non root user

pawels-optimizely:
## Summary
- run docker container as non root user
- tested just locally, it needs to be tested while deployed
- travis  handled this new user with no problem:

```
Step 11/13 : COPY --from=builder /etc/passwd /etc/passwd
 ---> e8bb295c9edf
Step 12/13 : USER agentuser
 ---> Running in 1035f68bb9ee
Removing intermediate container 1035f68bb9ee
 ---> 0ebd05595170
Step 13/13 : CMD ["/optimizely"]
 ---> Running in b140d42bd211
Removing intermediate container b140d42bd211
 ---> 2c13863c1264
[Warning] One or more build-args [NUM_RUNS SDK_BRANCH TESTAPP_BRANCH TRAVIS_COM_TOKEN] were not consumed
Successfully built 2c13863c1264
Successfully tagged agent-testapp:latest
```

## Issues
- https://optimizely.atlassian.net/browse/OASIS-6579

### DIFF
--- a/scripts/dockerfiles/Dockerfile.alpine
+++ b/scripts/dockerfiles/Dockerfile.alpine
@@ -1,6 +1,7 @@
 ARG GO_VERSION
 FROM golang:$GO_VERSION-alpine3.10 as builder
 # hadolint ignore=DL3018
+RUN addgroup -S 10001 && adduser -S agentuser -G 10001
 RUN apk add --no-cache make gcc libc-dev git curl
 WORKDIR /go/src/github.com/optimizely/agent
 COPY . .
@@ -9,4 +10,6 @@ RUN make setup build
 FROM alpine:3.10
 RUN apk add --no-cache ca-certificates
 COPY --from=builder /go/src/github.com/optimizely/agent/bin/optimizely /optimizely
+COPY --from=builder /etc/passwd /etc/passwd
+USER agentuser
 CMD ["/optimizely"]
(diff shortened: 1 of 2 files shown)


scripts/dockerfiles/Dockerfile.alpine
```diff
@@ -1,6 +1,7 @@
 ARG GO_VERSION
 FROM golang:$GO_VERSION-alpine3.10 as builder
 # hadolint ignore=DL3018
+RUN addgroup -S 10001 && adduser -S agentuser -G 10001
```
msohailhussain, line 4: I think This line should be at line 13, this is not running container.

scripts/dockerfiles/Dockerfile.static
```diff
@@ -1,5 +1,6 @@
 ARG GO_VERSION
 FROM golang:$GO_VERSION as builder
+RUN useradd -u 10001 agentuser
```
msohailhussain, line 3: any reason, not adding group here.

pawels-optimizely: group is not really needed but i will add it

scripts/dockerfiles/Dockerfile.alpine
pawels-optimizely: you need to add the user here and then get it from the builder

scripts/dockerfiles/Dockerfile.static
```diff
@@ -1,5 +1,7 @@
 ARG GO_VERSION
 FROM golang:$GO_VERSION as builder
+RUN  addgroup agentgroup
+RUN useradd -u 10001 agentuser -g agentgroup
```
deathweaselx86, line 4: try `addgroup -u 1000 agentgroup` and `useradd -u 1000 agentuser -g agentgroup`

pawels-optimizely: added as suggested

PR #239, opened 2020-06-27
feat: adding address as a configurable item.

pawels-optimizely:
## Summary
- Adding Host to Server config, defaulted to 0.0.0.0 to be accessible from outside container.  

https://optimizely.atlassian.net/browse/OASIS-6581


pawels-optimizely:
> LGTM. One request: can we add a test for setting a non-default address via config file or environment variable?

request addressed, added 1.2.3.4 IP as an address

mikeng13:
@msohailhussain  when you get a chance can you point Pawel to where in FSC we need to fix to those tests passing

PR #240, opened 2020-06-29
Introduce Agent interceptor plugins.

mikecdavis:
## Summary
- Adds support for middleware plugins via Interceptors
- Adds go-chi/httplog as an example plugin

This plugin system is meant to provide well-defined interfaces allowing extensions to be within their own namespaced packages. This provides functional separation from the core Optimizely Agent feature set with any number of integrations within the request flow.

Follow-up work would be to migrate some existing /pkg/middleware into /plugins/interceptors to serve as an example and to highlight the class of features that might constitute a plugin vs core functionality. Migration candidates include [auth.go](https://github.com/optimizely/agent/blob/master/pkg/middleware/auth.go), [metrics.go](https://github.com/optimizely/agent/blob/master/pkg/middleware/metrics.go), [request_header.go](https://github.com/optimizely/agent/blob/master/pkg/middleware/request_header.go).

The major con of this approach is that thrid-party plugins will require forking and building of Agent. Ideally new plugins can be generalized and contributed back into main.

### Open Questions
1. Should the Interceptor interface expose "routing" or should this be up to the MW handler to discern?
2. How do we distinguish middleware bound to a particular listener "all" vs "api" vs "admin" vs "webhook". Is this defined as part of the middleware or as part of configuration? (e.g. `server.interceptors`, `api.interceptors`, etc).

### Note
This plugin architecture is heavily modeled after [influxdata/telegraf](https://github.com/influxdata/telegraf/tree/master/plugins) plugin system. Here plugins are defined in their own package and registered with the global set of plugins.


plugins/middleware/registry.go
```diff
@@ -0,0 +1,15 @@
...
+type Creator func() Middleware
+
+var Middlewares = map[string]Creator{}
+
+type Middleware interface {
+	Handler() func(http.Handler) http.Handler
+}
+
+func Add(name string, creator Creator) {
+	Middlewares[name] = creator
```
mjc1283, line 14: Should we log an error or panic when there is already a `creator` set for `name`?

mikecdavis: Good idea. Probably `panic` makes the most sense to fail ASAP.

config/config.go
```diff
@@ -131,6 +134,7 @@ type ServerConfig struct {
 	KeyFile         string        `json:"keyFile"`
 	DisabledCiphers []string      `json:"disabledCiphers"`
 	HealthCheckPath string        `json:"healthCheckPath"`
+	Plugins         interface{}   `json:"plugins",mapstructure:"plugins"`
```
pawels-optimizely, line 137: should it be PluginConfigs ?

mikecdavis: I renamed to `Interceptors`.

README.md
```diff
@@ -72,17 +72,17 @@ Below is a comprehensive list of available configuration properties.
 
 |Property Name|Env Variable|Description|
 |---|---|---|
-|admin.auth.clients|N/A|Credentials for requesting access tokens. See: [Authorization Guide](./docs/auth.md)|
-|admin.auth.jwksURL|OPTIMIZELY_ADMIN_AUTH_JWKSURL|JWKS URL for validating access tokens. See: [Authorization Guide](./docs/auth.md)|
-|admin.auth.jwksUpdateInterval|OPTIMIZELY_ADMIN_AUTH_JWKSUPDATEINTERVAL|JWKS Update Interval for caching the keys in the background. See: [Authorization Guide](./docs/auth.md)|
-|admin.auth.hmacSecrets|OPTIMIZELY_ADMIN_AUTH_HMACSECRETS|Signing secret for issued access tokens. See: [Authorization Guide](./docs/auth.md)|
-|admin.auth.ttl|OPTIMIZELY_ADMIN_AUTH_TTL|Time-to-live of issued access tokens. See: [Authorization Guide](./docs/auth.md)|
+|admin.auth.clients|N/A|Credentials for requesting access tokens. See: [Authorization Guide](https://docs.developers.optimizely.com/full-stack/docs/authorization)|
```
mjc1283, line 75: I don't see any link from the main README to plugins documentation or plugins README - seems like we should have that.

pkg/server/server.go
```diff
@@ -104,6 +107,29 @@ func (s Server) Shutdown() {
 	}
 }
 
+func wrapHandler(handler http.Handler, conf config.PluginConfigs) http.Handler {
```
mjc1283, line 110: `wrapHandler` is not very descriptive, consider:
```suggestion
func wrapWithInterceptors(handler http.Handler, conf config.PluginConfigs) http.Handler {
```

mjc1283 (approved):
LGTM. Added comments & suggestions.

In our terminology are "interceptors" the same as "plugins"? Or maybe interceptors are a kind of plugin? If they're equivalent maybe we should get rid of "plugin" in variables/interfaces/log messages.

Are you thinking we should convert more of the candidate middlewares prior to releasing this?

pawels-optimizely (approved):
lgtm

docs/readme-sync/deploy-as-a-microservice/040 - configure-optimizely-agent/050 - agent-plugins.md
```diff
@@ -0,0 +1,88 @@
...
+hidden: false
+metadata:
+  title: "Agent plugins - Optimizely Full Stack"
+createdAt: "2020-09-21T20:30:00.000Z"
+updatedAt: "2020-09-21T20:30:00.000Z"
+---
+
+## Agent Plugins
+
+Optimizely Agent can be extended through the use of [plugins](https://docs.developers.optimizely.com/full-stack/docs/agent-plugins). Plugins are distinct from the standard Agent packages
```
fscelliott, line 14: This is a circular/self-referential link; delete.

fscelliott: Hey, by the way @mikecdavis  , I set up a preview in the travis stage so you can actually view the published agent-plugin page here before merge to master:

https://rollouts-sandbox-doc-test.readme.io/docs/agent-plugins

(I'm thinking of changing the sandbox domain, FYI)

fscelliott: Hey, by the way @mikecdavis  , I set up a preview in the travis stage so you can actually view the published agent-plugin page here before merge to master:

https://rollouts-sandbox-doc-test.readme.io/docs/agent-plugins

(I'm thinking of changing the sandbox domain, FYI...not sure you can see it now without being logged into ReadMe)

docs/readme-sync/deploy-as-a-microservice/040 - configure-optimizely-agent/050 - agent-plugins.md
```diff
@@ -0,0 +1,88 @@
...
+```go
+package all
+
+// Add imports here to trigger the plugin `init()` function
+import (
+    _ "github.com/optimizely/agent/plugins/interceptors/example"
+)
+```
+
+Enable the example interceptor by adding to `server.interceptors` within your `config.yaml`. Note that the yaml fields
```
fscelliott, line 79: also shows up as a weird line break; make 1 line

README.md
```diff
@@ -104,20 +104,20 @@ Below is a comprehensive list of available configuration properties.
...
+|server.batchRequests.maxConcurrency|OPTIMIZELY_SERVER_BATCHREQUESTS_MAXCONCURRENCY|Number of requests running in parallel. Default: 10|
+|server.batchRequests.operationsLimit|OPTIMIZELY_SERVER_BATCHREQUESTS_OPERATIONSLIMIT|Number of allowed operations. ( will flag an error if the number of operations exeeds this parameter) Default: 500|
 |server.disabledCiphers|OPTIMIZELY_SERVER_DISABLEDCIPHERS|List of TLS ciphers to disable when accepting HTTPS connections|
+|server.healthCheckPath|OPTIMIZELY_SERVER_HEALTHCHECKPATH|Path for the health status api. Default: /health|
 |server.host|OPTIMIZELY_SERVER_HOST|Host of server. Default: 127.0.0.1|
+|server.interceptors|N/A|Property used to enable and set [Interceptor](./plugins/interceptors/README.md) plugins|
```
fscelliott, line 112: seemed to redirect to 
https://github.com/optimizely/agent/blob/mikecdavis/plugin-arch/plugins/interceptors/README.md 
which was a 401...?

mikecdavis: Thanks, this should link to the readme.io doc.

mikecdavis:
@mjc1283  
> In our terminology are "interceptors" the same as "plugins"? Or maybe interceptors are a kind of plugin? If they're equivalent maybe we should get rid of "plugin" in variables/interfaces/log messages.

Right now they are 1-to-1, but we could have more distinct plugins in the future. (e.g. notifications listeners, user profile service backends)

> Are you thinking we should convert more of the candidate middlewares prior to releasing this?

It would be nice to have more examples prior to a release, but not a blocker.

PR #241, opened 2020-06-30
feat: Reject requests with invalid host

mjc1283:
## Summary

Introduce a new configuration property `server.allowedHosts`. When requests are received whose host doesn't match either one of these values, or the separately configured `server.host`, Agent rejects the request with a 404 response. The default value of `server.allowedHosts` is `["localhost"]`.

Includes code from [go-chi/hostrouter](https://github.com/go-chi/hostrouter) for determining request host based on `X-Forwarded-Host` header, `Forwarded` header, `Host` header, or host portion of URL.

## Test plan
- Manually tested running Agent locally
- New unit tests

## Issues
OASIS-6578


pkg/middleware/allowed_hosts.go
```diff
@@ -0,0 +1,116 @@
...
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			host := requestHost(r)
+			if allowedMap[host] {
+				next.ServeHTTP(w, r)
+				return
+			}
+			logger := GetLogger(r)
+			logger.Debug().Strs("allowedHosts", allowedHosts).Str("allowedPort", allowedPort).Str("Host", r.Host).Str("X-Forwarded-Host", r.Header.Get("X-Forwarded-Host")).Str("Forwarded", r.Header.Get("Forwarded")).Msg("Request failed allowed hosts check")
+			RenderError(errors.New("invalid request host"), http.StatusNotFound, w, r)
```
mikecdavis, line 47: Nit, but we can use a constant `error`.

pkg/middleware/allowed_hosts.go
```diff
@@ -0,0 +1,116 @@
...
+
+// AllowedHosts returns a middleware function that rejects requests whose host value does not match any host in allowedHosts.
+// Request host is determined in the following priority order:
+// 1. X-Forwarded-Host header value
+// 2. Forwarded header host= directive value
+// 3. Host property of request (see Host under https://golang.org/pkg/net/http/#Request)
+func AllowedHosts(allowedHosts []string, allowedPort string) func(next http.Handler) http.Handler {
+	allowedMap := make(map[string]bool)
+	for _, allowedHost := range allowedHosts {
+		allowedMap[fmt.Sprintf("%v:%v", allowedHost, allowedPort)] = true
```
mikecdavis, line 35: Is the port always included/required?

config.yaml
```diff
@@ -28,6 +28,15 @@ log:
...
 server:
+    ## List of allowed request host values.
+    ## Requests whose host value does not match one of these will be rejected with a 404 response.
+    ## Request host is determined in the following priority order:
+    ## 1. X-Forwarded-Host header value
+    ## 2. Forwarded header host= directive value
+    ## 3. Host property of request (see Host under https://golang.org/pkg/net/http/#Request)
+    allowedHosts:
+        - localhost
+        - 127.0.0.1
```
mikecdavis, line 39: Does this have to be enabled by default like this? Can we use a wildcard?

mikecdavis: BTW, this default is causing the FCS failures.

pkg/middleware/allowed_hosts.go
mjc1283: For determining the request host, it comes from X-Forwarded-Host header, Forwarded header, Host header, or the host portion of the URL. In all those places, port is optional. But, port is required in Agent configuration.

So we do have a problem here: if the user runs Agent on port 80 without https, or 443 with https, then they send requests omitting the port, this middleware will reject those. To solve I can add some additional handling here for those special cases.

pkg/middleware/allowed_hosts.go
```diff
@@ -0,0 +1,132 @@
...
+		// When appropriate, create entry in allowedMap without explicit port
+		if shouldAllowPortOmitted {
+			allowedMap[allowedHost] = true
+		}
+	}
+
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			host := requestHost(r)
+			if allowedMap[host] {
```
mikecdavis, line 57: This is an exact match which might be too restrictive. Looking at other libraries they tend to support wildcards and/or suffix matching e.g. `".example.com"` or `"*.example.com"`.

Examples:
- https://www.playframework.com/documentation/2.8.x/AllowedHostsFilter
- https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.hostfiltering.hostfilteringoptions.allowedhosts?view=aspnetcore-3.1
- https://docs.djangoproject.com/en/3.0/ref/settings/#allowed-hosts

PR #246, opened 2020-07-09
feat(auth-datafiles): Introducing support for authenticated datafiles

aliabbasrizvi:
**Changes:**
- Users can now optionally set `X-Optimizely-SDK-Key`  to pass in token for the datafile belonging to secure environment. In order to do so you can pass in the token separated from the SDK key by a colon. So, if SDK key is `my_key` and token is `my_token` then the header's value should be set to `my_key:my_token`. Similarly, you can set up the SDK key in the config to represent both the key and the token separated by a colon.
- SDK key regex updated to reflect this change.

### DIFF
--- a/api/openapi-spec/webhook.yaml
+++ b/api/openapi-spec/webhook.yaml
@@ -41,7 +41,6 @@ components:
           type: string
       required:
         - revision
-        - origin_url
         - cdn_url
     WebhookRequest:
       properties:
--- a/config/config.go
+++ b/config/config.go
@@ -77,7 +77,7 @@ func NewDefaultConfig() *AgentConfig {
 			DatafileURLTemplate: "https://cdn.optimizely.com/datafiles/%s.json",
 			EventURL:            "https://logx.optimizely.com/v1/events",
 			// https://github.com/google/re2/wiki/Syntax
-			SdkKeyRegex: "^\\w+$",
+			SdkKeyRegex: "^\\w+(:\\w+)?$",
 		},
 		Runtime: RuntimeConfig{
 			BlockProfileRate:     0, // 0 is disabled
--- a/pkg/middleware/cached.go
+++ b/pkg/middleware/cached.go
@@ -44,6 +44,9 @@ const OptlyExperimentKey = contextKey("experimentKey")
 // OptlySDKHeader is the header key for an ad-hoc SDK key
 const OptlySDKHeader = "X-Optimizely-SDK-Key"
 
+// OptlyDatafileAccessToken is the token used in requesting datafile for a secure environment
+const OptlyDatafileAccessToken = "X-Optimizely-Datafile-Access-Token"
+
 // CachedOptlyMiddleware implements OptlyMiddleware backed by a cache
 type CachedOptlyMiddleware struct {
 	Cache optimizely.Cache
@@ -60,7 +63,9 @@ func (mw *CachedOptlyMiddleware) ClientCtx(next http.Handler) http.Handler {
 			return
 		}
 
-		optlyClient, err := mw.Cache.GetClient(sdkKey)
+		datafileAccessToken := r.Header.Get(OptlyDatafileAccessToken)
+
+		optlyClient, err := mw.Cache.GetClient(sdkKey, datafileAccessToken)
 		if err != nil {
 			GetLogger(r).Error().Err(err).Msg("Initializing OptimizelyClient")
 
--- a/pkg/middleware/cached.go
+++ b/pkg/middleware/cached.go
@@ -44,9 +44,6 @@ const OptlyExperimentKey = contextKey("experimentKey")
 // OptlySDKHeader is the header key for an ad-hoc SDK key
 const OptlySDKHeader = "X-Optimizely-SDK-Key"
 
-// OptlyDatafileAccessToken is the token used in requesting datafile for a secure environment
-const OptlyDatafileAccessToken = "X-Optimizely-Datafile-Access-Token"
-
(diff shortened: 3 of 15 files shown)


api/openapi-spec/webhook.yaml
```diff
@@ -41,7 +41,6 @@ components:
           type: string
       required:
         - revision
-        - origin_url
```
aliabbasrizvi, line 44: We do not have `origin_url` in message for authenticated datafiles. Relaxing the schema.

pkg/optimizely/cache.go
```diff
@@ -95,7 +111,16 @@ func (c *OptlyCache) GetClient(sdkKey string) (*OptlyClient, error) {
 
 	// If we didn't "set" the key in this method execution then it was set in another thread.
 	// Recursively lookuping up the SDK key "should" only happen once.
-	return c.GetClient(sdkKey)
+	return c.GetClient(sdkKey, datafileAccessToken)
+}
+
+// UpdateConfigs is used to update config for all clients corresponding to a particular SDK key.
+func (c *OptlyCache) UpdateConfigs(sdkKey string) {
```
aliabbasrizvi, line 118: For a particular webhook message, it is possible that there are multiple clients with same SDK key, but separate tokens.

Since the client key for private datafiles looks like `sdk_key:token`, we update all the configs corresponding to clients that begin with the given SDK key. Let me know if this is acceptable.

pkg/middleware/cached.go
```diff
@@ -59,7 +62,9 @@ func (mw *CachedOptlyMiddleware) ClientCtx(next http.Handler) http.Handler {
 			return
 		}
 
-		optlyClient, err := mw.Cache.GetClient(sdkKey)
+		datafileAccessToken := r.Header.Get(OptlyDatafileAccessToken)
+
+		optlyClient, err := mw.Cache.GetClient(sdkKey, datafileAccessToken)
```
mjc1283, line 67: Imagine the following scenario:

1. Agent started with configuration: `datafileAccessTokens: {"mySdkKey": ["myToken1"]}`
2. Client added to cache with key: `"mySdkKey:myToken1"`
3. Request comes in with `X-Optimizely-Sdk-Key: mySdkKey`, and **no X-Optimizely-Datafile-Access-Token**

In this scenario, I think we want the existing client to be used. But this implementation will access the cache using only the sdk key as the cache key, not find anything, and then attempt to create a new client with SDK key only. If that SDK key corresponds to an authenticated datafile, it will fail to load.

Probably we want to treat the access token in the header as a "force" or "override" configuration - ignore any existing client we have for that SDK key, and create a new one using that SDK key and that access token.

pkg/optimizely/cache.go
```diff
@@ -95,7 +111,16 @@ func (c *OptlyCache) GetClient(sdkKey string) (*OptlyClient, error) {
...
+	for clientInfo := range c.optlyMap.IterBuffered() {
+		if strings.HasPrefix(clientInfo.Key, sdkKey) {
+			clientInfo.Val.(*OptlyClient).UpdateConfig()
```
mjc1283, line 121: Maybe to be safe, it could check the result of type cast attempt, log an error and return when it's unexpectedly not an `*OptlyClient`.

config/config.go
```diff
@@ -95,18 +95,19 @@ func NewDefaultConfig() *AgentConfig {
...
-	Log     LogConfig     `json:"log"`
-	Client  ClientConfig  `json:"client"`
-	Server  ServerConfig  `json:"server"`
-	Webhook WebhookConfig `json:"webhook"`
+	Version 				string 				`json:"version"`
+	Author  				string				`json:"author"`
+	Name    				string 				`json:"name"`
+
+	SDKKeys 				[]string 			`yaml:"sdkKeys" json:"sdkKeys"`
+	DatafileAccessTokens 	map[string][]string	`yaml:"datafileAccessTokens" json:"datafileAccessTokens"`
```
mikecdavis, line 103: nit, formatting got a little wonky here.

pkg/middleware/cached.go
mikecdavis: I thought we explicitly want the composite cache key behavior. Having redundant instances of Optly in the cache is ok since they are issued and revoked independently. Forcing an override when the access token changes can result in large number of cache reloads if two independent keys, from independent services are accessing the same Agent cluster.

pkg/optimizely/cache.go
```diff
@@ -107,14 +132,43 @@ func defaultLoader(
...
+			sdkKey = clientKey
+		}
+
+		// If there is a : then it is an authenticated datafile.
+		// First part is the sdkKey.
+		// Second part is the datafileAccessToken
+		if len(clientKeySplit) == 2 {
+			sdkKey = clientKeySplit[0]
+			datafileAccessToken = clientKeySplit[1]
+		}
```
mikecdavis, line 154: Per my previous comment, can we support this feature by allowing the access token to be passed via the existing header as opposed to adding a new header, just to combine it later?

pkg/middleware/cached.go
mjc1283: Yea that's a good point about forcing an override.

My main question is: will we require the datafile access token to be included in the request header, in order to use Agent with an authenticated datafile? If the answer is yes, this implementation works, otherwise it doesn't.

mjc1283: If we dont make datafile access tokens required in the request, then instead, they can be set in Agent configuration, Agent will fetch authenticated datafiles as needed, and requesters don't worry about it.

PR #247, opened 2020-07-14
feat(security): add SDK key validation configuration

mikecdavis:
## Summary
- Add validation parameter for SDK Key header `sdkKeyRegex` to client config
- Update cache loader to validate SDK Key before initializing the Optimizely client

This change allows the SDK Key found in the request header to be validated prior to making the templated URL fetch.

## Issues
- OASIS-6583

### DIFF
--- a/pkg/optimizely/cache.go
+++ b/pkg/optimizely/cache.go
@@ -105,6 +105,7 @@ func (c *OptlyCache) Wait() {
 	c.wg.Wait()
 }
 
+// ErrValidationFailure is returned when the provided SDK key fails initial validation
 var ErrValidationFailure = errors.New("sdkKey failed validation")
 
 func regexValidator(sdkKeyRegex string) func(string) bool {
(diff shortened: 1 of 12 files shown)


pkg/optimizely/cache.go
```diff
@@ -103,12 +105,32 @@ func (c *OptlyCache) Wait() {
 	c.wg.Wait()
 }
 
+var errValidationFailure = errors.New("sdkKey failed validation")
+
+func regexValidator(sdkKeyRegex string) func(sdkkey string) bool {
+	r, err := regexp.Compile(sdkKeyRegex)
+	if err != nil {
+		log.Fatal().Err(err).Msgf("invalid sdkKeyRegex configuration")
```
mikecdavis, line 113: This will cause Agent to `os.Exit(1)`
https://github.com/rs/zerolog/blob/9e51190d4708789500a087352d12d5f2e0c30b0d/log.go#L332-L338

PR #252, opened 2020-07-16
feat: block contenttype other than application/json

msohailhussain:
## Summary
-  Allow only application/json as content type for requests. Other content-type are not allowed.

## Test Plan
- Added unit tests

## Issues
OASIS-6582

### DIFF
--- a/pkg/routers/api.go
+++ b/pkg/routers/api.go
@@ -38,17 +38,18 @@ import (
 
 // APIOptions defines the configuration parameters for Router.
 type APIOptions struct {
-	maxConns        int
-	sdkMiddleware   func(next http.Handler) http.Handler
-	metricsRegistry *metrics.Registry
-	configHandler   http.HandlerFunc
-	activateHandler http.HandlerFunc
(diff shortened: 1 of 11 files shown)


msohailhussain:
@mjc1283 I know it's not needed but what about if later on admin allows binary stuff?

msohailhussain:
Based on our offlien discussion, we don't need configuration contentype, will remove that code.

pkg/routers/api.go
```diff
@@ -124,6 +127,7 @@ func WithAPIRouter(opt *APIOptions, r chi.Router) {
 	}
 
 	r.Use(middleware.SetTime)
+	r.Use(opt.allowedContentTypesMiddleware)
```
mjc1283, line 130: Why do we need to add `allowedContentTypesMiddleware` to `APIOptions`? Instead can we just do this here: 
```suggestion
	r.Use(chimw.AllowContentType("application/json"))
```

mjc1283:
@yasirfolio3 You can apply the content type checking middleware only to routes that accept `POST` method.

PR #264, opened 2020-08-14
[FSSDK-9409] Add gossip protocol for cluster membership.

mikecdavis:
## Summary
- Adds [hashicorp/memberlist](https://github.com/hashicorp/memberlist) dependency
- Creates "broadcasts" for config init/update and override set/remove
- Defines node "state" by loaded datafiles and current set of overrides
- Uses vendored `MapExperimentOverridesStore` from the go-sdk
- Adds top level admin index route
- Adds start_node script for creating a local clustered environment

Agent nodes within the same cluster should maintain some degree of consistency between them. Especially in the context of data file existence and revisions the 

## TODO
- Update the `MapExperimentOverridesStore` in the go-sdk for direct access to the underlying map
- Add unit tests


Tamara-Barum:
Internal ticket for tracking: FSSDK-9409

config/config.go
```diff
@@ -69,6 +69,10 @@ func NewDefaultConfig() *AgentConfig {
 			Pretty: false,
 			Level:  "info",
 		},
+		Cluster: ClusterConfig{
+			Port: 7896,
+			Host: "127.0.0.1",
```
pulak-opti, line 74: Forming cluster with localhost will not work in a distributed network isolated system like k8s pod

pulak-opti:
Agent itself does not hold any state. Datafile is modified outside the agent. So, forming a cluster is not necessary.

PR #271, opened 2020-08-26
fix: Allowed hosts middleware does not check port

mjc1283:
## Summary

This fixes an issue with the allowed hosts middleware that could prevent legitimate requests from being served.

The middleware extracts a host value from requests to compare agains the allow-list configured in `server.allowedHosts`. These request host values can take two possible forms (see: [Host header on MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host)):

1. Just a host (i.e. `example.com`), or
2. A host + port combination, separated by `:` (`example.com:8080`).

Previously, when the request host value did contain a port (the 2nd form), and that port differed from the `server.port` configuration value, the middleware would reject the request.

With this change, only the host is used for comparison against the configured allow-list. If the port is present, it is ignored.

## Test Plan
- Updated unit tests
- Manually tested


pkg/middleware/allowed_hosts.go
```diff
@@ -31,41 +32,32 @@ var errInvalidRequestHost = errors.New("invalid request host")
...
 	}
 
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			host := requestHost(r)
+
+			hostPortSplit := strings.Split(host, ":")
+			if len(hostPortSplit) == 2 {
+				host = hostPortSplit[0]
+			}
```
mikecdavis, line 48: Nit, but I think you can one line this:
```go
host := strings.Split(requestHost(r), ":")[0]
```

pkg/server/server_test.go
```diff
@@ -175,27 +178,11 @@ func TestNewServerHandlerAllowsValidHost(t *testing.T) {
...
-	req := httptest.NewRequest("GET", "http://example.com/v1/config", nil)
-	rec := httptest.NewRecorder()
 
+	req = httptest.NewRequest("GET", "http://example.com/v1/config", nil)
+	rec = httptest.NewRecorder()
 	srv.srv.Handler.ServeHTTP(rec, req)
-
 	assert.Equal(t, http.StatusOK, rec.Code)
-}
+}
```
mikecdavis, line 188: Nit, newline.

PR #275, opened 2020-09-04
feat: adding batching for agent

pawels-optimizely:
## Summary
- adding batching for agent
- it was done in middleware to take advantage of  ServeHTTP

### DIFF
--- a/pkg/middleware/batch.go
+++ b/pkg/middleware/batch.go
@@ -26,65 +26,74 @@ import (
 	"time"
 )
 
-// BatchResposeItem holds the structure for each item
-type BatchResposeItem struct {
+// BatchResponse has the structure for the final response
+type BatchResponse struct {
+	StartedAt     time.Time           `json:"startedAt"`
+	EndedAt       time.Time           `json:"endedAt"`
+	ErrorCount    int                 `json:"errorCount"`
+	ResponseItems []ResponseCollector `json:"response"`
+}
+
+// NewBatchResponse constructs a BatchResponse with default values
+func NewBatchResponse() *BatchResponse {
+	return &BatchResponse{
+		StartedAt:     time.Now(),
+		ResponseItems: make([]ResponseCollector, 0),
+	}
+}
+
+func (br *BatchResponse) append(col ResponseCollector) {
+	if col.Status != http.StatusOK {
+		br.ErrorCount++
+	}
+
+	br.ResponseItems = append(br.ResponseItems, col)
+	br.EndedAt = time.Now()
+}
+
+// ResponseCollector collects responses for the writer
+type ResponseCollector struct {
 	Status    int         `json:"status"`
 	RequestID string      `json:"requestID"`
 	Method    string      `json:"method"`
 	URL       string      `json:"url"`
 	Body      interface{} `json:"body"`
-}
 
-// BatchResponse has the structure for the final response
-type BatchResponse struct {
-	StartedAt     time.Time          `json:"startedAt"`
-	EndedAt       time.Time          `json:"endedAt"`
-	ErrorCount    int                `json:"errorCount"`
-	ResponseItems []BatchResposeItem `json:"response"`
-}
+	StartedAt time.Time `json:"startedAt"`
+	EndedAt   time.Time `json:"endedAt"`
 
-// BatchWriter implements http.ResponseWriter
-type BatchWriter struct {
-	BResponse BatchResponse
-}
-
-func (br *BatchWriter) append(rec *ResponseCollector) {
-	br.BResponse.EndedAt = time.Now()
-	if rec.BResponse.Status != http.StatusOK {
-		br.BResponse.ErrorCount++
-	}
-
-	br.BResponse.ResponseItems = append(br.BResponse.ResponseItems, rec.BResponse)
-	br.BResponse.EndedAt = time.Now()
-}
-
-// ResponseCollector collects responses for the writer
-type ResponseCollector struct {
 	headerMap http.Header
+}
 
-	BResponse BatchResposeItem
+// NewResponseCollector constructs a ResponseCollector with default values
+func NewResponseCollector(op BatchOperation) ResponseCollector {
+	return ResponseCollector{
+		headerMap: make(http.Header),
+		Method:    op.Method,
+		URL:       op.URL,
+		StartedAt: time.Now(),
+		Status:    http.StatusOK,
+	}
 }
 
 // WriteHeader sets the status code
 func (rec *ResponseCollector) WriteHeader(code int) {
-	rec.BResponse.Status = code
+	rec.Status = code
 }
 
 // Write is just the collector for BatchResponse
 func (rec *ResponseCollector) Write(b []byte) (int, error) {
-
 	var data interface{}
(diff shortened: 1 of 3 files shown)


pkg/middleware/batch.go
```diff
@@ -0,0 +1,150 @@
...
+	"time"
+)
+
+// BatchResposeItem holds the structure for each item
+type BatchResposeItem struct {
+	Status       int         `json:"status"`
+	RequestID    string      `json:"requestID"`
+	Method       string      `json:"method"`
+	URL          string      `json:"url"`
+	OrigResponse interface{} `json:"origResponse"`
```
mikecdavis, line 35: Perhaps "body" is more natural.

pkg/middleware/batch.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// BatchResponse has the structure for the final response
+type BatchResponse struct {
+	StartedAt  time.Time          `json:"startedAt"`
+	EndedAt    time.Time          `json:"endedAt"`
+	ErrorCount int                `json:"errorCount"`
+	Response   []BatchResposeItem `json:"response"`
```
mikecdavis, line 43: Should probably be plural like: "responseItems"

pkg/middleware/batch.go
```diff
@@ -0,0 +1,150 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,        *
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *
+ * See the License for the specific language governing permissions and      *
+ * limitations under the License.                                           *
+ ***************************************************************************/
+
+// Package middleware //
+package middleware
+
+import (
```
mjc1283, line 20: Suggest: follow import convention used in other files - separate into these groups, in this order, separated by newline:
1. Go stdlib packages
2. Optimizely packages (agent & Go SDK)
3. 3rd party packages.

pkg/middleware/batch.go
```diff
@@ -0,0 +1,150 @@
...
+		Body        map[string]interface{} `json:"body"`
+		Params      map[string]string      `json:"params"`
+		Headers     map[string]string      `json:"headers"`
+	} `json:"operations"`
+}
+
+// BatchRouter intercepts requests for the given url to return a StatusOK.
+func BatchRouter(next http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == "POST" && strings.HasSuffix(strings.ToLower(r.URL.Path), "/batch") {
```
mjc1283, line 102: This is a separate custom implementing of routing. Why not define a route the usual way?

pkg/middleware/batch.go
```diff
@@ -0,0 +1,150 @@
...
+
+			decoder := json.NewDecoder(r.Body)
+			var req Request
+			err := decoder.Decode(&req)
+			if err != nil {
+				render.JSON(w, r, `{"error": "cannot decode the operation body"}`)
+				return
+			}
```
mjc1283, line 110: Don't we also need to set an error status (400)?

pkg/middleware/batch.go
```diff
@@ -0,0 +1,150 @@
...
+	return 0, nil
+}
+
+// Header returns header map
+func (rec *BatchWriter) Header() http.Header {
+	return rec.headerMap
+}
+
+// Request is the original request that is used for batching
+type Request struct {
```
mjc1283, line 88: `Request` could be confused with the http package `Request`.
```suggestion
type BatchRequest struct {
```

pkg/middleware/batch.go
```diff
@@ -0,0 +1,150 @@
...
+	statusCode int
+	headerMap  http.Header
+	method     string
+	url        string
+
+	BResponse BatchResponse
+}
+
+// WriteHeader sets the status code
+func (rec *BatchWriter) WriteHeader(code int) {
```
mjc1283, line 57: Is this method only used in tests?

pkg/middleware/batch.go
```diff
@@ -0,0 +1,150 @@
...
+
+			decoder := json.NewDecoder(r.Body)
+			var req Request
+			err := decoder.Decode(&req)
+			if err != nil {
+				render.JSON(w, r, `{"error": "cannot decode the operation body"}`)
+				return
+			}
+			batchWriter := BatchWriter{200, make(http.Header), "", "",
+				BatchResponse{StartedAt: time.Now(), Response: []BatchResposeItem{}}}
```
mikecdavis, line 112: Having a single global `BatchWriter` like this is going to be problematic and prevent us from extending to add concurrency support. Since we're reusing the statusCode, method, url and headers they can easily collide from one request to another. Alternatively we should construct a _new_ writer (collector) for each operation and _append_ the contents onto the BatchResponse instance.

The pseudo code would be something like:
```go
br := BatchResponse{StartedAt: time.Now(), Response: []BatchResposeItem{}}}
for _, op := range req.Operations {
	// Construct collector
	col := ResponseCollector{StartedAt: time.Now()}

	// Build request
	bytesBody, e := json.Marshal(op.Body)
	...

	// Make request
	next.ServeHTTP(col, opReq)
	col.EndedAt = time.Now()

	// Append response item
	br.append(col)
}
br.EndedAt = time.Now()
render.JSON(w, r, br)
```

pkg/middleware/batch.go
```diff
@@ -0,0 +1,150 @@
...
+
+			for _, op := range req.Operations {
+				bytesBody, e := json.Marshal(op.Body)
+				if err != nil {
+					GetLogger(r).Error().Err(e).Msg("cannot convert operation body to bytes for operation id " + op.OperationID)
+					continue
+				}
+				reader := bytes.NewReader(bytesBody)
+				newReq, e := http.NewRequest(op.Method, op.URL, reader)
```
mikecdavis, line 121: Nit, but lets name this `opReq` for clarity.

mikecdavis (changes requested):
We need to separate out the concerns of collecting the individual operation responses with the building of the batch response.

mjc1283 (commented):
I am not sure about implementing our own `ResponseWriter`. Some of the comments in the [documentation of ResponseWriter](https://golang.org/pkg/net/http/#ResponseWriter) seem like they're describing the built-in implementation of that interface:

For example:
> // If WriteHeader has not yet been called, Write calls
    // WriteHeader(http.StatusOK) before writing the data. If the Header
    // does not contain a Content-Type line, Write adds a Content-Type set
    // to the result of passing the initial 512 bytes of written data to
    // DetectContentType.

Is our implementation of `Write` doing that? Do we need to do that? When it's not clearly documented, I'm not sure what the responsibilities of the implementations are.

pkg/middleware/batch.go
pawels-optimizely: you need to define this method to overwrite the original WriteHeader()

pkg/middleware/batch.go
pawels-optimizely: Then we would need to use http.Client() to make a direct call to the request. Using ServeHTTP is easier and it feels more natural

mikecdavis (approved):
LGTM

PR #276, opened 2020-09-04
feat(allowed hosts): Allow matching subdomains or all hosts

mjc1283:
## Summary

Add support for matching subdomains, or matching any host, in the `server.allowedHosts` configuration property:
- `.example.com` matches `my.example.com`, `hello.world.example.com`, etc.
- `.` matches all hosts

## Issues
OASIS-7084


mikecdavis (approved):

pawels-optimizely (commented):
lgtm

PR #277, opened 2020-09-09
fix: Removed vulnerable version coreos/etcd

msohailhussain:
## Summary
- coreos/etc  was vulnerable and  dependency link is changed. Replaced with that link.

## Issues
OASIS-6869


mikecdavis (approved):
LGTM, if you can remove some of the previous excludes that would be great!!!

PR #278, opened 2020-09-09
feat: batching in parallel

pawels-optimizely:
## Summary
- making batching requests in parallel
- the order of batching responses is not guaranteed.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -93,6 +93,10 @@ func NewDefaultConfig() *AgentConfig {
 			KeyFile:         "",
 			DisabledCiphers: make([]string, 0),
 			Host:            "127.0.0.1",
+			BatchRequests: BatchRequestsConfig{
+				ParallelRequests: 1,
+				OperationsLimit:  3,
+			},
 		},
 		Webhook: WebhookConfig{
 			Port: "8085",
@@ -159,14 +163,15 @@ type LogConfig struct {
 
 // ServerConfig holds the global http server configs
 type ServerConfig struct {
-	AllowedHosts    []string      `json:"allowedHosts"`
-	ReadTimeout     time.Duration `json:"readTimeout"`
-	WriteTimeout    time.Duration `json:"writeTimeout"`
-	CertFile        string        `json:"certFile"`
-	KeyFile         string        `json:"keyFile"`
-	DisabledCiphers []string      `json:"disabledCiphers"`
-	HealthCheckPath string        `json:"healthCheckPath"`
-	Host            string        `json:"host"`
+	AllowedHosts    []string            `json:"allowedHosts"`
+	ReadTimeout     time.Duration       `json:"readTimeout"`
+	WriteTimeout    time.Duration       `json:"writeTimeout"`
+	CertFile        string              `json:"certFile"`
+	KeyFile         string              `json:"keyFile"`
+	DisabledCiphers []string            `json:"disabledCiphers"`
+	HealthCheckPath string              `json:"healthCheckPath"`
+	Host            string              `json:"host"`
+	BatchRequests   BatchRequestsConfig `json:"batchRequests"`
 }
 
 func (sc *ServerConfig) isHTTPSEnabled() bool {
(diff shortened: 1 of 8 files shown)


mikecdavis (changes requested):
This is looking great. I think we need a few more safeguards like the using the errgroup context along with setting both a batch size limit and the number of concurrent requests.

pkg/middleware/batch.go
```diff
@@ -132,40 +175,20 @@ func BatchRouter(next http.Handler) http.Handler {
 			}
 
 			batchRes := NewBatchResponse()
-			for _, op := range req.Operations {
+			var eg errgroup.Group
```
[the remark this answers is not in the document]
pawels-optimizely: Unfortunately, it won't work for our case. We cannot have request's context in async calls. As soon as our controller is done with the request, it will be canceled. Technically, as soon as our client got response , then it can close the connection. It won't work for our parallel processing, especially when you have less workers than the number of operations

pkg/middleware/batch.go
```diff
@@ -117,59 +128,86 @@ type BatchRequest struct {
...
-				batchRes.append(col)
-			}
+				ch := make(chan struct{}, batchRequests.ParallelRequests)
 
-			render.JSON(w, r, batchRes)
-			return
+				for _, op := range req.Operations {
+					op := op
+					eg.Go(func() error {
+						ch <- struct{}{}
```
mikecdavis, line 193: This approach starts _all_ the the goroutines, then blocks their execution. We should probably block the goroutine from being submitted until there is an available lease.

config/config.go
```diff
@@ -189,6 +194,12 @@ type APIConfig struct {
 	EnableOverrides     bool              `json:"enableOverrides"`
 }
 
+// BatchRequestsConfig holds the configuration for batching
+type BatchRequestsConfig struct {
+	ParallelRequests int `json:"parallelRequests"`
```
mikecdavis, line 199: I think something like "maxConcurrency" is more self explanatory.

pkg/middleware/batch.go
```diff
@@ -117,59 +128,86 @@ type BatchRequest struct {
...
+					return
 				}
 
-				for paramKey, paramValue := range op.Params {
-					values := opReq.URL.Query()
-					values.Add(paramKey, paramValue)
-					opReq.URL.RawQuery = values.Encode()
+				if len(req.Operations) > batchRequests.OperationsLimit {
+					GetLogger(r).Info().Msg(fmt.Sprintf("Too many operations, setting to the first %d operations", batchRequests.OperationsLimit))
+					req.Operations = req.Operations[:batchRequests.OperationsLimit]
```
mikecdavis, line 182: We should fail the entire request with a `4XX` (perhaps [422 Unprocessable Entity](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422)). Returning a successful partial response would be very confusing and difficult to handle from a client consumer.

mikecdavis (changes requested):

pkg/middleware/batch.go
pawels-optimizely: ok, then the client can adjust

pkg/middleware/batch.go
pawels-optimizely: yeah, starting all goroutines at once is cheap (go can handle a mln of go routines started at once, since they are green threads) , and I think it will be faster that way instead of starting goroutines in batches.

mikecdavis: Hm, ok. I think what makes this work in this scenario is that we're blocking on a channel. So the go scheduler knows it can put this goroutine into a waiting state. If the goroutine was blocked on some other system call, then I think this would have adverse implications. I'll probably do some research on this to satisfy my own understanding.

pawels-optimizely: after a much thought, I am changing it

pkg/middleware/batch.go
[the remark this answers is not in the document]
pawels-optimizely: we have resolved it and agreed what is a hard error vs soft error

mikecdavis (approved):
LGTM. just need to propagate the error from the `errgroup` through the response.

PR #279, opened 2020-09-21
1st stab at API reference for batch

fscelliott:
## Summary
- The "what"; a concise description of each logical change
- Another change

The "why", or other context.

## Issues

### DIFF
--- a/api/openapi-spec/openapi.yaml
+++ b/api/openapi-spec/openapi.yaml
@@ -433,7 +433,6 @@ components:
           enum:
             - GET
             - POST
-            - wrong_method
         url:
           type: string
         body:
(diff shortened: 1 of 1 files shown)


api/openapi-spec/openapi.yaml
```diff
@@ -340,6 +369,52 @@ components:
...
+              requestID:
+                type: string
+              operationID:
+                type: string
+              method:
+                type: string
+                enum:
+                  - GET
+                  - POST
+                  - wrong_method
```
pawels-optimizely, line 409: I think this should be removed, we can only allow GET and POST

pawels-optimizely (changes requested):

pawels-optimizely (approved):
lgtm

PR #286, opened 2020-10-09
Mpirnovar/batch endpoint tests

Mat001:
## Summary
- added acceptance tests for the /v1/batch endpoint
- tests added: valid response for single operation, valid response for multiple operations with different SDK keys, invalid response 400, invalid response for exceeded number of operations

## Issues
- Feature Jira ticket [OASIS-7204](https://optimizely.atlassian.net/browse/OASIS-7204)


.travis.yml
```diff
@@ -76,7 +76,8 @@ jobs:
       install:
         - pip install -r tests/acceptance/requirements.txt
       script:
-        - pytest -vv --diff-type=split tests/acceptance/test_acceptance/ --host http://localhost:8080
+        - pytest -vv --diff-type=split tests/acceptance/test_acceptance/ --host http://localhost:8080 -k "not test_batch_422"
+        - pytest -vv --diff-type=split tests/acceptance/test_acceptance/ --host http://localhost:8080 -k "test_batch_422"
```
mikecdavis, line 80: Why do we need two different runs?

mikecdavis (approved):
Overall LGMT, but can we consolidate to a single run of the test suite. It doesn't appear to actually need two distinct runs.

.travis.yml
pawels-optimizely: I think one is setting some extra env variables, to get 422 return code (Unprocessable Entity)

pawels-optimizely (approved):
lgtm

PR #288, opened 2020-10-13
Logging improvement

deepthawtz:
## Summary
- Stop logging SDK keys. As many others also do we collect logs for search and don't want these keys to be viewable
- Log incoming API requests. Yes, there is the /metrics endpoint but we want to also see what reqs the agent is handling via logs using our existing tools.

I'm open to suggestions, e.g., if we only want to make the http logger optional (via config).

### DIFF
--- a/pkg/middleware/utils.go
+++ b/pkg/middleware/utils.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
-	"strings"
 
 	"github.com/optimizely/agent/pkg/optimizely"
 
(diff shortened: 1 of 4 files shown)


deepthawtz:
CI tests failing at

```
$ ./scripts/pull_travis_ci_tools.sh
~/travisci-tools ~/build/optimizely/agent
Initialized empty Git repository in /home/travis/travisci-tools/.git/
remote: Repository not found.
fatal: Authentication failed for 'https://@github.com/optimizely/travisci-tools.git/'
```

PR #293, opened 2021-01-07
feat: Adding new /datafile endpoint

yasirfolio3:
## Summary
- Providing datafile access through `/datafile` endpoint

## Issues
- https://optimizely.atlassian.net/browse/OASIS-5500

### DIFF
--- a/api/openapi-spec/openapi.yaml
+++ b/api/openapi-spec/openapi.yaml
@@ -40,7 +40,7 @@ paths:
             application/json:
               schema:
                 description: Optimizely Datafile
-                $ref: '#/components/schemas/Datafile'
+                type: object
         '401':
           description: Unauthorized, invalid JWT
         '403':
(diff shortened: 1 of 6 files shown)


api/openapi-spec/openapi.yaml
```diff
@@ -28,6 +28,23 @@ paths:
...
+      operationId: getDatafile
+      description: Returns the json datafile for the given environment.
+      responses:
+        '200':
+          description: Valid response
+          content:
+            application/json:
+              schema:
+                description: Optimizely Datafile
+                type: object
```
msohailhussain, line 43: should it be object or string?

msohailhussain (approved):
have one comment, otherwise looks good.

msohailhussain:
Before merging, need to check with FSC.

api/openapi-spec/openapi.yaml
yasirfolio3: We are returning datafile as json hence have written type as `object` here. If we intend to return datafile as `string` then we will have to map it against a key map to render it as json in response.

pkg/handlers/get_datafile_test.go
```diff
@@ -0,0 +1,81 @@
+package handlers
```
mjc1283, line 1: Add license header

mjc1283 (approved):
LGTM. Please add license header to get_datafile_test.go.

PR #309, opened 2021-05-12
fix: disabling readmesync for now

msohailhussain:
## Summary
- For now disabling readme sync, will rely on manual push to readme.


jaeopt (approved):
LGTM

PR #310, opened 2021-06-09
Enforce go fmt in agent.

keppel2:
## Summary
- Fix and enforce Go fmt. Do not simplify.

## Test plan
- Run lint with fmt on and simplify off.


msohailhussain (changes requested):
please update command.

msohailhussain (approved):
lgtm

PR #316, opened 2021-08-02
Add events, attributes, and audiences to agent.

keppel2:
Agent changes for v2 config.


keppel2:
Not sure why FSC isn't reporting back. I'll point it back to master on go-sdk and see what's going on.

PR #322, opened 2021-09-06
feat(Config-v2): Support for optlyconfig-v2 added.

yasirfolio3:
## Summary
- Support for OptimizelyConfig v2 added.

## Issues
- All Config-v2 tests should pass.


go.mod
```diff
@@ -12,7 +12,7 @@ require (
 	github.com/go-kit/kit v0.9.0
 	github.com/google/uuid v1.1.1
 	github.com/lestrrat-go/jwx v0.9.0
-	github.com/optimizely/go-sdk v1.6.2-0.20210507191537-aba050a09690
+	github.com/optimizely/go-sdk v1.6.2-0.20210903184112-9d97c16f5787
```
msohailhussain, line 15: this should be update with the go release.

msohailhussain (commented):
looks fine to me. leaving it on matjaz for acceptance and one more review from jae/jacob

api/openapi-spec/openapi.yaml
```diff
@@ -306,23 +306,57 @@ components:
...
     Feature:
       properties:
         id:
           type: string
         key:
           type: string
+        experimentRules:
+          type: array
+          items:
+            $ref: '#/components/schemas/Experiment'
```
The-inside-man, line 351: Should these be OptimizelyExperiments or does it follow the same as Experiment from config? Same for feature and attribute.

tests/acceptance/test_acceptance/test_batch.py
```diff
@@ -47,36 +47,78 @@ def test_batch_valid_reponse(session_obj):
...
+					"key": "15439080355",
+					"variationsMap": {
+						"15404730339": {
+							"featureEnabled": true,
+							"id": "15404730339",
+							"key": "15404730339",
+							"variablesMap": {
+								"fff": {
+									"id": "15427520260",
+									"key": "fff",
```
Mat001, line 94: Just curious, did you have to add these variables in the app.optmizely.com project or were they already there? (I created the project but I forgot)

tests/acceptance/test_acceptance/test_config.py
```diff
@@ -7,121 +7,322 @@
...
-                },
-                "variation_2": {
-                    "id": "16927770169",
-                    "key": "variation_2",
-                    "featureEnabled": false,
-                    "variablesMap": {}
-                }
-            }
+  "environmentKey": "",
+  "sdkKey": "",
```
Mat001, line 11: Hmm, here we're getting empty values. I know why, it's because it's already an existing project and the flag wasn't turned on. I wonder if we should modernize our project for Agent acceptance tests. Currently it uses legacy UI and doesn't include sdk and env keys.
maybe I can create a Jira ticket. And then we update this test so we will have sdk and environment key values in.
What do you think @msohailhussain

Mat001: As a minimum this test works, the keys show, but with empty values. At least the keys can be verified.

Mat001 (approved):
LGTM. Can you please respond about my comment about updating the app.optmizely project for acceptance tests. thx

api/openapi-spec/openapi.yaml
```diff
@@ -470,7 +526,7 @@ components:
         type:
           type: string
         value:
-          type: object
+          type: string
```
jaeopt, line 529: why this change? JSON type?

jaeopt (changes requested):
Clarification and suggestions for some naming changes

tests/acceptance/test_acceptance/test_batch.py
yasirfolio3: they were already there.

api/openapi-spec/openapi.yaml
yasirfolio3: It is of type `string` in go-sdk. Guess it was wrong earlier. https://github.com/optimizely/go-sdk/blob/master/pkg/config/optimizely_config.go#L106

go.mod
yasirfolio3: reverting it will cause the unit tests to fail.

api/openapi-spec/openapi.yaml
```diff
@@ -315,7 +315,7 @@ components:
       required:
         - id
         - key
-    Attribute:
+    OptimizelyAttribute:
```
jaeopt, line 318: It looks ok but want to confirm that theres is no implications of these renaming to existing agent users.

jaeopt (changes requested):
Looks good. One more name change suggested.

api/openapi-spec/openapi.yaml
yasirfolio3: Since we have only updated the model names and haven't removed/renamed any old `key/value` pair in the optimizelyConfig, there shouldn't be any implication i believe. What do you think? @msohailhussain

jaeopt (approved):
LGTM

jaeopt (approved):
LGTM

PR #326, opened 2021-12-24
feat(ups): Added UserProfileService Support.

yasirfolio3:
## Summary
- Provided 2 out of the box user profile services including `in-memory`. and `redis`.
- Added `lifo` and `fifo` ordering in `in-memory` UPS.
- Added support for custom user profile services

## Build Size
- Without UPS: 12.1 mb
- With UPS: 12.7 mb

## Build Time
- Without UPS: 7.384s
- With UPS: 7.387s

## Performance
- Without UPS: 1000 decide all calls in 580ms
- With UPS (disabled by default): 1000 decide all calls in 580ms

## UserProfileService EndPoints

1. To lookup a user profile, use agent's `POST /v1/lookup`:

```
In the request `application/json` body, include the `userId`. The full request looks like this:

```curl
curl --location --request POST 'http://localhost:8080/v1/lookup' \
--header 'X-Optimizely-SDK-Key: YOUR_SDK_KEY' \
--header 'Accept: text/event-stream' \
--header 'Content-Type: application/json' \
--data-raw '{
  "userId": "string"
}'
```

2. To save a user profile, use agent's `POST /v1/save`:

```
In the request `application/json` body, include the `userId` and `experimentBucketMap`. The full request looks like this:

```curl
curl --location --request POST 'http://localhost:8080/v1/save' \
--header 'X-Optimizely-SDK-Key: YOUR_SDK_KEY' \
--header 'Accept: text/event-stream' \
--header 'Content-Type: application/json' \
--data-raw '{
    "userId": "string",
    "experimentBucketMap": {
        "experiment_id_to_save": {
            "variation_id": "variation_id_to_save"
        }
    }
}'
```

## Out of Box UserProfileService Usage

1. To use the in-memory `UserProfileService`, update the `config.yaml` as shown below:
```
## configure optional User profile service
userProfileService:
      default: "in-memory"
      services:
        in-memory: 
          ## 0 means no limit on capacity
          capacity: 0
          ## supports lifo/fifo
          storageStrategy: "fifo"
```
2. To use the redis `UserProfileService`, update the `config.yaml` as shown below:
```
## configure optional User profile service
userProfileService:
      default: "redis"
      services:
        redis: 
          host: "your_host"
          password: "your_password"
          database: 0 ## your database
```
3. To use the rest `UserProfileService`, update the `config.yaml` as shown below:
```
## configure optional User profile service
userProfileService:
      default: "rest"
      services:
        rest:
          host: "your_host"
          lookupPath: "/lookup_endpoint"
          lookupMethod: "POST"
          savePath: "/save_endpoint"
          saveMethod: "POST"
          userIDKey: "user_id"
          headers: 
            "header_key": "header_value"
```

Implement 2 `POST` api's `/lookup_endpoint` and `/save_endpoint` on your `host`. Api methods will be `POST` by default but can be  updated through `lookupMethod` and `saveMethod` properties. Similarly, request parameter key for `user_id` can also be updated  using `userIDKey` property.
    
- `lookup_endpoint` should accept `user_id` in its json body or query (depending upon the method type) and if successful, return the status code `200` with json response (keep in mind that when sending response, `user_id` should be substituted with value of `userIDKey` from config.yaml):

```
{
  "experiment_bucket_map": {
    "saved_experiment_id": {
      "variation_id": "saved_variation_id"
    }
  },
  "user_id": "saved_user_id"
}
```
- `save_endpoint` should accept the following parameters in its json body or query (depending upon the method type) and return the status code `200` if successful (keep in mind that `user_id` should be substituted with the value of`userIDKey` from config.yaml):   

```
{
  "experiment_bucket_map": {
    "experiment_id_to_save": {
      "variation_id": "variation_id_to_save"
    }
  },
  "user_id": "user_id_to_save"
}
```

## Custom UserProfileService Implementation

To implement a custom user profile service, followings steps need to be taken:
1. Create a struct that implements the `decision.UserProfileService` interface in `plugins/userprofileservice/services`.
2. Add a `init` method inside your UserProfileService file as shown below:
```
func init() {
	myUPSCreator := func() decision.UserProfileService {
		return &yourUserProfileServiceStruct{
		}
	}
	userprofileservice.Add("my_ups_name", myUPSCreator)
}
```
3. Update the `config.yaml` file with your `UserProfileService` config as shown below:

```
## configure optional User profile service
userProfileServices:
    default: "my_ups_name"
    services:
        my_ups_name: 
           ## Add those parameters here that need to be mapped to the UserProfileService
           ## For example, if the UPS struct has a json mappable property called `host`
           ## it can updated with value `abc.com` as shown
           host: “abc.com”
```
- If a user has created multiple `UserProfileServices` and wants to override the `default` `UserProfileService` for a specifc `sdkKey`, they can do so by providing the `UserProfileService` name in the request Header `X-Optimizely-UPS-Name`.

## Tests
- Unit tests added.
- FSC should pass.


pkg/optimizely/cache.go
```diff
@@ -114,6 +119,11 @@ func (c *OptlyCache) UpdateConfigs(sdkKey string) {
 	}
 }
 
+// SetUserProfileService sets maps userProfileService against the given sdkKey
+func (c *OptlyCache) SetUserProfileService(sdkKey, userProfileService string) {
```
msohailhussain, line 123: Please add lock here.

yasirfolio3: we are using `userProfileServiceMap cmap.ConcurrentMap` here which handles concurrency itself.

anvth:
Hey team, this is amazing news and something my team was considering contributing! 💟 

Just a question, in the example config we see there is an option to specify custom URL for User Profile Service. However, reviewing PR I could not see an implementation for that. Is there a plan to support that any time soon? Even the PR description does not cover this.

Thank you :-)

yasirfolio3:
> Hey team, this is amazing news and something my team was considering contributing! 💟
> 
> Just a question, in the example config we see there is an option to specify custom URL for User Profile Service. However, reviewing PR I could not see an implementation for that. Is there a plan to support that any time soon? Even the PR description does not cover this.
> 
> Thank you :-)

Hi @anvth , We have updated the PR description, please have a look and let us know if there's anything else we can help with, Thanks.

pkg/optimizely/cache.go
The-inside-man: I think we should still add the mutex here. For reading the ConcurrentMap should be fine but when setting, I believe the mutex should be used to ensure no race condition, even with the ConcurrentMap, just to be safe.. I could be wrong though.

The-inside-man (approved):
LGTM! Just the one comment around mutex as @msohailhussain suggested.

204Constie:
Hey team, this is amazing news for us as we were planning to implement this. 😸 
A significant question here that we would have would be about the custom UPS implementation. Contrary to how in memory and redis are implemented there is no way of implementing it by simply providing certain options in the config file. 
Did you consider an option to make it less flexible but not requiring changes in the code ? 
What about having an out of the box option for microservice UPS ?

thank you 😸

plugins/userprofileservice/services/in_memory_ups.go
```diff
@@ -0,0 +1,121 @@
...
+				delete(u.ProfilesMap, oldProfile)
+			}
+
+			// Push new entry to ordered list
+			switch u.StorageStrategy {
+			case "lifo":
+				u.lifoOrderedProfiles = append(u.lifoOrderedProfiles, profile.ID)
+			default:
+				// fifo by default
+				u.fifoOrderedProfiles <- profile.ID
```
jaeopt, line 108: The benefit of supporting fifo/lifo is not clear to me. Can we just implement the default LRU cache unless we see clear reasons they prefer fifo/lifo?

plugins/userprofileservice/services/rest_ups.go
```diff
@@ -0,0 +1,115 @@
...
+	Host       string            `json:"host"`
+	Headers    map[string]string `json:"headers"`
+	LookupPath string            `json:"lookupPath"`
+	SavePath   string            `json:"savePath"`
+}
+
+// Lookup is used to retrieve past bucketing decisions for users
+func (r *RestUserProfileService) Lookup(userID string) (profile decision.UserProfile) {
+	if userID == "" {
+		return
```
jaeopt, line 45: Is it ok not returning profile in go?

jaeopt (changes requested):
I have a suggestion and a clarification. 
Other than those, it looks great to me as far as I can follow with go :)

plugins/userprofileservice/services/rest_ups.go
yasirfolio3: yes, if we don't do it here, it will just return empty profile struct.

204Constie:
hello! 😸 all of the updates to rest UPS look wonderful 😻 all configurable via config.yml file 😻 great work!
i have a small question. i can read that both `lookup` and `save` methods expect the rest endpoints to return the profile on `success`. this is for sure necessary in case of `lookup` but is it really required in case of `save` method ? does it use the response of the `save` method or expect to use it in the future ? or it actually only require the response of status code `2XX` and would in truth behave the same in case of `200` and `204` ?

yasirfolio3:
> hello! 😸 all of the updates to rest UPS look wonderful 😻 all configurable via config.yml file 😻 great work! i have a small question. i can read that both `lookup` and `save` methods expect the rest endpoints to return the profile on `success`. this is for sure necessary in case of `lookup` but is it really required in case of `save` method ? does it use the response of the `save` method or expect to use it in the future ? or it actually only require the response of status code `2XX` and would in truth behave the same in case of `200` and `204` ?

Hi @204Constie , In case of `save`, only 200 status code needs to be sent if successful (204 is not supported). The documentation is actually stating the payload required for the `save` request  👍

204Constie:
> > hello! 😸 all of the updates to rest UPS look wonderful 😻 all configurable via config.yml file 😻 great work! i have a small question. i can read that both `lookup` and `save` methods expect the rest endpoints to return the profile on `success`. this is for sure necessary in case of `lookup` but is it really required in case of `save` method ? does it use the response of the `save` method or expect to use it in the future ? or it actually only require the response of status code `2XX` and would in truth behave the same in case of `200` and `204` ?
> 
> Hi @204Constie , In case of `save`, only 200 status code needs to be sent if successful (204 is not supported). The documentation is actually stating the payload required for the `save` request 👍

😸 this is perfectly fine, i was just wondering why is this the case. is the response used in any way ? is there some other reason ? i'm simply curious

yasirfolio3:
> > > hello! 😸 all of the updates to rest UPS look wonderful 😻 all configurable via config.yml file 😻 great work! i have a small question. i can read that both `lookup` and `save` methods expect the rest endpoints to return the profile on `success`. this is for sure necessary in case of `lookup` but is it really required in case of `save` method ? does it use the response of the `save` method or expect to use it in the future ? or it actually only require the response of status code `2XX` and would in truth behave the same in case of `200` and `204` ?
> > 
> > 
> > Hi @204Constie , In case of `save`, only 200 status code needs to be sent if successful (204 is not supported). The documentation is actually stating the payload required for the `save` request 👍
> 
> 😸 this is perfectly fine, i was just wondering why is this the case. is the response used in any way ? is there some other reason ? i'm simply curious

So for `save` API, no response is required. Even if a response is sent, it will be ignored by agent. Agent only checks for status code `200`.

pkg/optimizely/cache.go
```diff
@@ -192,12 +203,63 @@ func defaultLoader(
...
+		}
+
+		optimizelyClient, err := optimizelyFactory.Client(
+			clientOptions...,
 		)
+		return &OptlyClient{optimizelyClient, configManager, forcedVariations, clientUserProfileService}, err
+	}
+}
+
+func getUserProfileService(sdkKey string, userProfileServiceMap cmap.ConcurrentMap, conf config.ClientConfig) decision.UserProfileService {
```
msohailhussain, line 225: add description

pkg/optimizely/cache.go
```diff
@@ -192,12 +203,63 @@ func defaultLoader(
...
+
+	createAndReturnUPSWithName := func(upsName string) decision.UserProfileService {
+		if clientConfigUPSMap, ok := conf.UserProfileService["services"].(map[string]interface{}); ok {
+			if defaultUserProfileServiceMap, ok := clientConfigUPSMap[upsName].(map[string]interface{}); ok {
```
msohailhussain, line 229: userprofileserviceconfig should be better name.

pkg/optimizely/cache.go
```diff
@@ -192,12 +203,63 @@ func defaultLoader(
...
+							return upsInstance
+						}
+					}
+				}
+			}
+		}
+		return nil
+	}
+
+	// Check if ups type was provided in the request headers
```
msohailhussain, line 253: name

pkg/optimizely/cache.go
```diff
@@ -192,12 +203,63 @@ func defaultLoader(
...
+	if ups, ok := userProfileServiceMap.Get(sdkKey); ok {
+		if upsNameStr, ok := ups.(string); ok && upsNameStr != "" {
+			return createAndReturnUPSWithName(upsNameStr)
```
msohailhussain, line 256: initialize or New

pkg/optimizely/interface.go
```diff
@@ -25,6 +25,7 @@ import (
 type Cache interface {
 	GetClient(sdkKey string) (*OptlyClient, error)
 	UpdateConfigs(sdkKey string)
+	SetUserProfileService(sdkKey, userProfileService string)
```
msohailhussain, line 28: add description, why we need to add here.

plugins/userprofileservice/services/in_memory_ups.go
```diff
@@ -0,0 +1,124 @@
...
+// InMemoryUserProfileService represents the in-memory implementation of UserProfileService interface
+type InMemoryUserProfileService struct {
+	Capacity int `json:"capacity"`
+	// StorageStrategy defines the storage strategy. Supported values include fifo and lifo.
+	StorageStrategy     string `json:"storageStrategy"`
+	ProfilesMap         map[string]decision.UserProfile
+	fifoOrderedProfiles chan string
+	lifoOrderedProfiles []string
+	lock                sync.RWMutex
+	isReady             bool
```
msohailhussain, line 39: why we need isReady?

msohailhussain (changes requested):
please address

README.md
```diff
@@ -95,6 +95,7 @@ Below is a comprehensive list of available configuration properties.
 |client.pollingInterval|OPTIMIZELY_CLIENT_POLLINGINTERVAL|The time between successive polls for updated project configuration. Default: 1m|
 |client.queueSize|OPTIMIZELY_CLIENT_QUEUESIZE|The max number of events pending dispatch. Default: 1000|
 |client.sdkKeyRegex|OPTIMIZELY_CLIENT_SDKKEYREGEX|Regex to validate SDK keys provided in request header. Default: ^\\w+(:\\w+)?$|
+|client.userProfileService|OPTIMIZELY_CLIENT_USERPROFILESERVICE| Property used to enable and set UserProfileServices. Default: ./config.yaml|
```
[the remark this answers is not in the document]
yasirfolio3: copied from interceptors: https://github.com/optimizely/agent/pull/326/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L110

config/config.go
```diff
@@ -78,6 +78,10 @@ func NewDefaultConfig() *AgentConfig {
 			EventURL:            "https://logx.optimizely.com/v1/events",
 			// https://github.com/google/re2/wiki/Syntax
 			SdkKeyRegex: "^\\w+(:\\w+)?$",
+			UserProfileService: UserProfileServiceConfigs{
```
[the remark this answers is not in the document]
yasirfolio3: Suggested by linter, Golang follows the same convention for initialising new properties.

go.sum
```diff
@@ -85,7 +110,18 @@ github.com/modern-go/concurrent v0.0.0-20180228061459-e0a39a4cb421/go.mod h1:6dJ
...
+github.com/nxadm/tail v1.4.8/go.mod h1:+ncqLTQzXmGhMZNUePPaPqPvBxHAIsmXswZKocGu+AU=
 github.com/oklog/ulid v1.3.1/go.mod h1:CirwcVhetQ6Lv90oh/F+FBtV6XMibvdAFo93nm5qn4U=
+github.com/onsi/ginkgo v1.6.0/go.mod h1:lLunBs/Ym6LB5Z9jYTR76FiuTmxDTDusOGeTQH+WWjE=
+github.com/onsi/ginkgo v1.12.1/go.mod h1:zj2OWP4+oCPe1qIXoGWkgMRwljMUYCdkwsT2108oapk=
+github.com/onsi/ginkgo v1.16.4 h1:29JGrr5oVBm5ulCWet69zQkzWipVXIol6ygQUe/EzNc=
+github.com/onsi/ginkgo v1.16.4/go.mod h1:dX+/inL/fNMqNlz0e9LfyB9TswhZpCVdJM/Z6Vvnwo0=
+github.com/onsi/gomega v1.7.1/go.mod h1:XdKZgCCFLUoM/7CFJVPcG8C1xQ1AJ0vpAezJrB7JYyY=
+github.com/onsi/gomega v1.10.1/go.mod h1:iN09h71vgCQne3DLsj+A5owkum+a2tYe+TOCB1ybHNo=
+github.com/onsi/gomega v1.16.0 h1:6gjqkI8iiRHMvdccRJM8rVKjCWk6ZIm6FTm3ddIe4/c=
+github.com/onsi/gomega v1.16.0/go.mod h1:HnhC7FXeEQY45zxNK3PPoIUhzk/80Xly9PcubAlGdZY=
```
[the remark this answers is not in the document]
yasirfolio3: Have updated the PR description with build size with redis.

pkg/handlers/lookup.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+type lookupBody struct {
+	UserID string `json:"userId"`
+}
+
+// UPSResponseOut defines the UPS API response
+type UPSResponseOut struct {
+	UserID              string                 `json:"userId"`
+	ExperimentBucketMap map[string]interface{} `json:"experimentBucketMap"`
```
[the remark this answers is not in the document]
yasirfolio3: This property is just used inside this file to send response.

pkg/handlers/save_test.go
```diff
@@ -0,0 +1,240 @@
...
+			UserID: "testUser",
+		},
+	}
+
+	expected := convertToUserProfile(body)
+	actual := suite.oc.UserProfileService.Lookup("testUser")
+	suite.Equal(expected, actual)
+}
+
+func (suite *SaveTestSuite) TestSave() {
```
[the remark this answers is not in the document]
yasirfolio3: `save` only checks for `userId` to be non-empty, in which case it directly saves the userprofile against the `userId` in the map. There's no other validation to it. This can be verified from the following test: https://github.com/optimizely/agent/pull/326/files#diff-495a600fac0bb299d45a5c0b0c193c4d3ce65eff2f38123d927f4ced74a2d0a9R195

plugins/userprofileservice/services/in_memory_ups.go
yasirfolio3: since we need to initialize `fifoOrderedProfiles` with `capacity` and this can only be done when first save is called, `isReady` is used to this purpose. All other required initialisations also take place if `isReady` is false.

`u.fifoOrderedProfiles = make(chan string, u.Capacity)`

msohailhussain (approved):
lgtm

jaeopt (approved):
LGTM

PR #329, opened 2022-01-14
feat(allowed hosts): Added support to remove sdkKey from logs.

yasirfolio3:
## Summary
- `sdkKey` can be removed from logs by setting `log.includeSdkKey` configuration property to `false`.

### DIFF
--- a/pkg/middleware/utils.go
+++ b/pkg/middleware/utils.go
@@ -1,5 +1,5 @@
 /****************************************************************************
- * Copyright 2019, Optimizely, Inc. and contributors                        *
+ * Copyright 2019,2022 Optimizely, Inc. and contributors                    *
  *                                                                          *
  * Licensed under the Apache License, Version 2.0 (the "License");          *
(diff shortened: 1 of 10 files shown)


pkg/middleware/utils.go
```diff
@@ -49,11 +49,14 @@ func GetOptlyClient(r *http.Request) (*optimizely.OptlyClient, error) {
 
 // GetLogger gets the logger with some info coming from http request
 func GetLogger(r *http.Request) *zerolog.Logger {
-	sdkKey := r.Header.Get(OptlySDKHeader)
 	reqID := r.Header.Get(OptlyRequestHeader)
+	logger := log.With().Str("requestId", reqID).Logger()
```
msohailhussain, line 53: bring it after `if`, stay with the same sequence.

msohailhussain (approved):
lgtm

pkg/middleware/utils.go
msohailhussain: nvm

PR #331, opened 2022-02-09
feat: UPS in memory acceptance tests

ozayr-zaviar:
## Summary
- Testcase for in-memory storage and save and lookup APIs

## Tests
- All other acceptance tests and FSC tests should pass.


tests/acceptance/test_acceptance/test_ups.py
```diff
@@ -0,0 +1,148 @@
...
+        "attributes": {
+          "attr_1": "hola"
+        }
+      },
+      "reasons": []
+    }
+"""
+
+
+@pytest.mark.parametrize(
```
Mat001, line 42: You could probably not use parameterization because you only test one case.
initially I used parameterization to combine several similar scenarios together, but here I think it would be cleaner if you just pass parameters into the funct signature

tests/acceptance/test_acceptance/test_ups.py
```diff
@@ -0,0 +1,148 @@
...
+    for _ in range(5):
+      resp = create_and_validate_request_and_response(ENDPOINT_DECIDE, 'post', session_obj, payload=payload,
+                                                      params=params)
+
+      assert json.loads(expected_response) == resp.json()
+      assert resp.status_code == expected_status_code, resp.text
+      resp.raise_for_status()
+
+
+@pytest.mark.parametrize(
```
Mat001, line 77: Same here. No parameterization needed.
Ps: Unless you are expecting to add more test variations later

tests/acceptance/test_acceptance/test_ups.py
```diff
@@ -0,0 +1,148 @@
...
+    "flag_key, expected_response, expected_status_code", [
+        ("feature_2", expected_flag_key, 200)
+    ],
+    ids=["valid case"])
+def test_ups__feature(session_obj, flag_key, expected_response, expected_status_code):
+    """
+    Test validates:
+    Same decide is returned after calling multiple times.
```
Mat001, line 50: Maybe change to "Same decide call...".

tests/acceptance/test_acceptance/test_ups.py
```diff
@@ -0,0 +1,148 @@
...
+    resp.raise_for_status()
+    assert json.loads(lookup_response) == resp.json()
+
+    params = {"keys": 'feature_2'}
+    resp = create_and_validate_request_and_response(ENDPOINT_DECIDE, 'post', session_obj, payload=decide_payload,
+                                                    params=params)
+
+    assert json.loads(expected_after_saved_variation) == resp.json()
+    assert resp.status_code == expected_status_code, resp.text
+    resp.raise_for_status()
```
Mat001, line 148: Could we add a test for invalid payload. For example that UPS doesn't save it.
Maybe one test for invalid payload for each of the two tests?

Mat001 (changes requested):

Mat001 (approved):
lgtm

jaeopt (changes requested):
I see "experimentBucketMap" and "userId" in openapi spec unlike our standard keys as "experiment_bucket_map" and "user_id". Can you check if this is an error in this spec only?

yasirfolio3:
> 

This is intentional since we have been following camel case in other api's as-well eg `decide`.

jaeopt (approved):
LGTM

tests/acceptance/test_acceptance/test_decide.py
```diff
@@ -49,7 +49,7 @@
           "attr_1": "hola"
         }
       },
-      "reasons": ["Audiences for experiment feature_2_test collectively evaluated to true."]
+      "reasons": ["User \"matjaz\" was previously bucketed into variation \"variation_1\" of experiment \"feature_2_test\"."]
```
Mat001, line 52: Where does this reason and subsequent reasons below come from?
Because this s the core decide api test, not UPS test

Mat001 (changes requested):

tests/acceptance/test_acceptance/test_decide.py
yasirfolio3: These are returned with the `decide` api response. The following changes were required because of the recent changes in the datafile used by these tests.

Mat001: @yasirfolio3 Hmm. I know, I made the changes to the datafile :). I mean I fixed the failing tests.
But my acceptance tests locally pass without these changes.

Mat001: Let me see the TravisCi run of these tests

Mat001: My acc tests pass on the code without UPS. Maybe you changed smth in the UPS that requires this adjustment?

Mat001: I see. The tests pass on master without UPS (that's the same as my local run). 
They start failing in your PR with UPS. Let's see if your PR changed something such that now decide response is different?

Mat001 (changes requested):

tests/acceptance/test_acceptance/test_decide.py
yasirfolio3: https://github.com/optimizely/agent/pull/331/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R94
The following change is causing agent to use `in-memory` UPS by default which is causing sticky bucketing hence the reasons have changed to `User \"matjaz\" was previously bucketed into variation \"variation_1\" of experiment \"feature_2_test\`.

Mat001: OK, great.
That's fine.
But I think we'll have a problem deploying ACC tests on Fargate where we'll need to disable UPS acceptance tests.

Mat001: Hmm, let me think. Actually I think we'll be fine.

On fargate we run acc tests like so:
pytest -vv -rA --diff-symbols tests/acceptance/test_acceptance -k 'not test_overrides.py and not test_batch_422' --host $TARGET_HOST

if I add a "not" condition for the two UPS tests, but the OPTIMIZELY_CLIENT_USERPROFILESERVICE is still set, then OPTIMIZELY_CLIENT_USERPROFILESERVICE changes decide responses fr other tests correct?

yasirfolio3: yes, because of sticky bucketing, OPTIMIZELY_CLIENT_USERPROFILESERVICE will return the same variation.

Mat001: ok

Mat001 (approved):

PR #334, opened 2022-03-08
fix: JWT Security Warning

yasirfolio3:
## Summary
- Updating JWT library from `https://github.com/dgrijalva/jwt-go` to `https://github.com/golang-jwt/jwt` because of security issues since the former one is no longer maintained and it is suggested to move to the later one.
- Updating to version `[v3.2.1](https://github.com/golang-jwt/jwt/releases/tag/v3.2.1)` of the library since post this version, only go version 1.15 and upwards are supported whereas agent uses go `1.13`.

## Issues
- https://github.com/advisories/GHSA-w73w-5m7g-f7qc

## GO Release Policy
- Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release. We fix critical problems, including [critical security problems](https://go.dev/security), in supported releases as needed by issuing minor revisions (for example, Go 1.6.1, Go 1.6.2, and so on).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,8 @@ language: shell
 env:
   global:
     # GO VERSION is set here with GIMME_GO_VERSION see: https://github.com/travis-ci/gimme
-    # may also want to run `go mod edit -go=1.15` to fix go.mod as well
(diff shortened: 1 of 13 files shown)


msohailhussain (approved):
lgtm

PR #340, opened 2022-04-04
feat(datafile-cache): Add support for datafile cache service.

msohailhussain:
## Summary
- Add support for datafile cache service which allows nodes in a cluster to re-use cached datafile from redis instance.
- Fix context for user profile service.

## Tests
- All Unit and FSC tests should pass.

## Redis Datafile Cache Service
To enable redis datafile cache service, Update the `config.yaml` file with your `datafileCacheService` config as shown below:

```
## Datafile cache to save initial datafile fetch call in a multi-node environment.
datafileCacheService:
      enabled: true
      services:
        redis: 
          host: "your_redis_host_address"
          password: "your_password"
          database: 0
```

### DIFF
--- a/cmd/optimizely/testdata/default.yaml
+++ b/cmd/optimizely/testdata/default.yaml
@@ -33,9 +33,11 @@ client:
   eventURL: "https://logx.localhost.com/v1"
   sdkKeyRegex: "custom-regex"
   datafileCacheService:
-    redis: 
-      host: "localhost:6379"
-      password: "123"
+    enabled: true
+    services:
+      redis: 
+        host: "localhost:6379"
+        password: "123"
   userProfileService:
     default: "in-memory"
     services:
--- a/cmd/optimizely/testdata/default.yaml
+++ b/cmd/optimizely/testdata/default.yaml
@@ -32,6 +32,10 @@ client:
   datafileURLTemplate: "https://localhost/v1/%s.json"
   eventURL: "https://logx.localhost.com/v1"
   sdkKeyRegex: "custom-regex"
+  datafileCacheService:
+    redis: 
+      host: "localhost:6379"
+      password: "123"
   userProfileService:
     default: "in-memory"
     services:
--- a/pkg/optimizely/cache.go
+++ b/pkg/optimizely/cache.go
@@ -201,10 +201,13 @@ func defaultLoader(
 				Password: redisCacheConfig.Password,
 				DB:       redisCacheConfig.Database,
 			})
-			if datafile, err := redisClient.Get(ctx, sdkKey).Result(); err != nil && datafile != "" {
+			datafile, err := redisClient.Get(ctx, sdkKey).Result()
+			if err == nil && datafile != "" {
 				// Set datafile in config manager so it uses the cached datafile for initialization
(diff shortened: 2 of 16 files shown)


cmd/optimizely/testdata/default.yaml
```diff
@@ -32,6 +32,10 @@ client:
   datafileURLTemplate: "https://localhost/v1/%s.json"
   eventURL: "https://logx.localhost.com/v1"
   sdkKeyRegex: "custom-regex"
+  datafileCacheService:
```
msohailhussain, line 35: datafileCacheServices will be more relevant.

msohailhussain, line 35: add one more property enableDatafileCacheService and based on that, execute the logic.

pkg/optimizely/cache.go
```diff
@@ -186,25 +195,39 @@ func defaultLoader(
 			log.Info().Msg(message)
 		}
 
+		// Options for PollingProjectConfigManager
+		options := []sdkconfig.OptionFunc{}
+
+		// Check if datafile is already present in cache
+		cachedDatafile, cacheService := getDatafileFromCacheService(ctx, sdkKey, conf)
```
msohailhussain, line 202: add a condition of cache is enabled then go into this method.

pkg/optimizely/cache.go
```diff
@@ -226,7 +249,7 @@ func defaultLoader(
 		}
 
 		var clientUserProfileService decision.UserProfileService
-		if clientUserProfileService = getUserProfileService(sdkKey, userProfileServiceMap, conf); clientUserProfileService != nil {
+		if clientUserProfileService = getUserProfileService(ctx, sdkKey, userProfileServiceMap, conf); clientUserProfileService != nil {
```
msohailhussain, line 252: Why we need to change this method.

pkg/optimizely/datafilecacheservice/redis_cache_service.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+// DatafileCacheService represents interface for datafileCacheService
+type DatafileCacheService interface {
+	GetDatafileFromCacheService(ctx context.Context, sdkKey string) string
+	SetDatafileInCacheService(ctx context.Context, sdkKey, datafile string)
+}
+
+// RedisCacheService represents the redis implementation of DatafileCacheService interface
+type RedisCacheService struct {
```
msohailhussain, line 34: this struct is same for UPS as well, can't we use the same.

cmd/optimizely/testdata/default.yaml
yasirfolio3: This is to keep it similar to the naming convention we are following for `userProfileService`.

pkg/optimizely/cache.go
yasirfolio3: Earlier, We were giving `UPS` a different context altogether which might've resulted in `UPS` requests still running for a little while after agent had shut down. This fix is to pass the original agent context to `UPS` so that once agents shuts down, `UPS` terminates all its requests as well.

PR #341, opened 2022-04-08
WIP: add flag to detect if UPS is on or off in accept tests

Mat001:
## Summary
- We want to have UPS enabled or disabled within Acceptance tests. Or rather wew want to detect if UPS is used or not and run acceptance tests accordingly.
- The reason for this is that we have a discrepancy: Agent repo is on v2.7.0 that has UPS active, while sidedoor-teraform repo that runs tests on fargate runs on Agent v2.6.0 without UPS. But THEY BOTH USE THE SAME ACC TESTS.
- Fargate is currently not configured to deal w UPS that's why we are running pre-ups version of agent there.
- We need flexibility inside acceptance tests to handle cases when UPS is on or off. 
- This PR uses function url_points_to_cluster() to check if host URL in pytest command points to AWS cluster URL. If so, then it run tests that don't support UPS. There are only a few such cases. For example tests for decide would normally fail in Agent versions where UPS is not supported, because UPS affects these decide tests. To have decide tests running on non-UPS versions of Agent like on AWS clusters we have added this "flag" function. If host points to defined AWS URLs then expected response doesn't include sticky variation from UPS, but it expects normal variation.
- 

## Issues
- ["OASIS-8268"](https://optimizely.atlassian.net/browse/OASIS-8268)


Mat001:
PS:
1. Local Agent correctly uses the UPS version of the tests.
2. Fargate Agent correctly uses the non-UPS version of the tests (https://app.travis-ci.com/github/optimizely/sidedoor-terraform/builds/249397187)

msohailhussain (approved):
lgtm

PR #347, opened 2022-12-05
[FSSDK-9412] add expose prometheus through custom collector

204Constie:
## Summary
dirty PR that presents the idea of exposing metrics to prometheus via the use of custom prometheus collector

an implementation of the collector interface https://github.com/prometheus/client_golang/blob/main/prometheus/collector.go#L27 
similar to this collector: https://github.com/prometheus/client_golang/blob/main/prometheus/expvar_collector.go
 it is not an ideal solution as the documentation states ( https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/collectors#NewExpvarCollector ) 
```NewExpvarCollector returns a newly allocated expvar Collector.
An expvar Collector collects metrics from the expvar interface. It provides a quick way to expose numeric values that are already exported via expvar as Prometheus metrics. Note that the data models of expvar and Prometheus are fundamentally different, and that the expvar Collector is inherently slower than native Prometheus metrics. Thus, the expvar Collector is probably great for experiments and prototying, but you should seriously consider a more direct implementation of Prometheus metrics for monitoring production systems.```
however it does not require implementation of any generic instrumentation tooling like opentelemetry


## Issues


Tamara-Barum:
Internal ticket for tracking: FSSDK-9412

pulak-opti:
Hi @204Constie 
Thanks for the PR. Could you please update your branch with the master branch?

pulak-opti:
Hi @204Constie 
Thanks for the idea about collecting Prometheus metrics. However, the feature is added here https://github.com/optimizely/agent/pull/348. Please reach out if you have anything.

PR #349, opened 2022-12-13
make ups save rest process not blocking activate

204Constie:
## Summary
makes the save process on UPS non blocking on the /activate endpoint
does this only at the level of REST UPS 

it just conveys the idea. the PR would create goroutine per function invocation and does not provide error reporting on failed save


msohailhussain:
@204Constie Can you please close this PR in favor of this PR. https://github.com/optimizely/agent/pull/350
We'll merge soon and ship it.

204Constie:
closing in favour of https://github.com/optimizely/agent/pull/350

PR #360, opened 2023-02-28
Fix: build was not triggering on pushing tags

mnoman09:
## Summary
- fixed the regex to validate the tags value


msohailhussain (approved):
lgtm

PR #361, opened 2023-02-28
fix: tags push doesn't work with regex

msohailhussain:
## Summary
- Fixed tags push doesn't work with regex.

## Issues


yasirfolio3 (approved):
LGTM

PR #362, opened 2023-02-28
fix: Adding back travis.yml

yasirfolio3:
## Summary
- Adding back travis since some jobs were failing on GitAction.


msohailhussain (approved):
lgtm

PR #364, opened 2023-03-02
[FSSDK-8971] feat: add FetchQualifiedSegments to Decide

andrewleap-optimizely:
## Summary
- FetchSegments and FetchSegmentsOptions added to the Decide endpoint.
- If FetchSegments is true, FetchQualifiedSegments is called from within Decide.

## Ticket
[FSSDK-8971](https://jira.sso.episerver.net/browse/FSSDK-8971)


go.mod
```diff
@@ -12,7 +12,7 @@ require (
 	github.com/golang-jwt/jwt/v4 v4.5.0
 	github.com/google/uuid v1.3.0
 	github.com/lestrrat-go/jwx v0.9.0
-	github.com/optimizely/go-sdk v1.8.3
+	github.com/optimizely/go-sdk v1.8.4-0.20230216074708-27b2772ccf33
```
yasirfolio3, line 15: This should be done using `go mod tidy` and `go mod clean` to automatically update `go.sum` and remove irrelevant indirect dependencies which were needed by go-sdk `1.8.3`.

pkg/handlers/decide.go
```diff
@@ -18,23 +18,27 @@
...
-	UserID          string                 `json:"userId"`
-	UserAttributes  map[string]interface{} `json:"userAttributes"`
-	DecideOptions   []string               `json:"decideOptions"`
-	ForcedDecisions []ForcedDecision       `json:"forcedDecisions,omitempty"`
+	UserID               string                            `json:"userId"`
+	UserAttributes       map[string]interface{}            `json:"userAttributes"`
+	DecideOptions        []string                          `json:"decideOptions"`
+	ForcedDecisions      []ForcedDecision                  `json:"forcedDecisions,omitempty"`
+	FetchSegments        bool                              `json:"fetchSegments"`
+	FetchSegmentsOptions []segment.OptimizelySegmentOption `json:"fetchSegmentsOptions,omitempty"`
```
yasirfolio3, line 41: Can we confirm that if user sends a wrong value here then this will not break? maybe have 3 unit tests:
1. pass a single incorrect value in array
2. pass a mixture of correct and incorrect values
3. pass a empty array

yasirfolio3 (changes requested):
looks good, just some minor changes required.

jaeopt (changes requested):
Looks great!  
Can we update openapi.yml  (new decide request params, 5xx error response on fetechSegments failure)?

yasirfolio3 (approved):
LGTM, just needs an update to `openapi.yaml` as @jaeopt suggested.

jaeopt (approved):
LGTM

PR #365, opened 2023-03-03
[FSSDK-8838] feat(odp-cache): Adds support for odp cache.

yasirfolio3:
## Summary
-  Provides 2 out of the box ODP cache including `in-memory` and `redis`.
- `in-memory` will use the builtin LRU cache with user provided `size` and `timeout`.
-  Adds support for custom odp cache.
-  If no odp cache is provided, agent will use go-sdk's in built LRU Cache with default `size` and `timeout`.

## Out of Box ODP Cache Usage

1. To use the in-memory `ODPCache`, update the `config.yaml` as shown below:
```
## configure optional ODP Cache
odp:
  segmentsCache:
    default: "in-memory"
    services:
      in-memory: 
        ## 0 means cache will be disabled
        size: 0
        ## timeout after which the cached item will become invalid.
        ## 0 means records will never be deleted
        timeout: 0s
```

2. To use the redis `ODPCache`, update the `config.yaml` as shown below:
```
## configure optional ODP Cache
odp:
  segmentsCache:
    default: "redis"
    services:
      redis: 
        host: "your_host"
        password: "your_password"
        database: 0 ## your database
        timeout: 0s
```

## Custom ODPCache Implementation

To implement a custom odp cache, followings steps need to be taken:
1. Create a struct that implements the `cache.Cache` interface in `plugins/odpcache/services`.
2. Add a `init` method inside your ODPCache file as shown below:
```
func init() {
	myCacheCreator := func() cache.Cache {
		return &yourCacheStruct{
		}
	}
	odpcache.Add("my_cache_name", myCacheCreator)
}
```
3. Update the `config.yaml` file with your `ODPCache` config as shown below:

```
## configure optional ODPCache
odp
  segmentsCache:
    default: "my_cache_name"
    services:
      my_cache_name: 
        ## Add those parameters here that need to be mapped to the ODPCache
        ## For example, if the ODP struct has a json mappable property called `host`
        ## it can updated with value `abc.com` as shown
        host: “abc.com”
```
- If a user has created multiple `ODPCache` services and wants to override the `default` `ODPCache` for a specific `sdkKey`, they can do so by providing the `ODPCache` name in the request Header `X-Optimizely-ODP-Cache-Name`.

- Whenever a request is made with a unique `sdkKey`, The agent node handling that request creates and caches a new `ODPCache`. To keep the `ODPCache` type consistent among all nodes in a cluster, it is recommended to send the request Header `X-Optimizely-ODP-Cache-Name` in every request made.

## Tests
- Unit tests added.
- FSC should pass.

## Ticket
[FSSDK-8838](https://jira.sso.episerver.net/browse/FSSDK-8838)


config.yaml
```diff
@@ -157,6 +157,31 @@ client:
...
+      segmentsRequestTimeout: 10s
+      ## This will override all other settings
+      segmentsCache:
+        default: ""
+        services:
+        #   in-memory: 
+        #     size: 0
+        #     timeout: 0 # seconds
+        #   redis: 
+        #     host: "localhost:6379"
```
jaeopt, line 181: Don't we need to control redis timeout as well? We can reuse `segmentsCacheTimeout` but it may be confusing since we do not use `segmentsCacheSize` for redis.

plugins/odpcache/services/redis_cache.go
```diff
@@ -0,0 +1,111 @@
...
+		Addr:     r.Address,
+		Password: r.Password,
+		DB:       r.Database,
+	})
+}
+
+func init() {
+	redisCacheCreator := func() cache.Cache {
+		return &RedisCache{
+			Expiration: 0 * time.Second,
```
jaeopt, line 107: We need timeout control for redis cache. Can't we add `timeout` to config?

jaeopt (changes requested):
Looks great! Some suggestions to reuse go-sdk cache. Let's discuss options.

cmd/optimizely/main.go
```diff
@@ -93,6 +95,11 @@ func loadConfig(v *viper.Viper) *config.AgentConfig {
 		conf.Client.UserProfileService = userProfileService
 	}
 
+	// Check if JSON string was set using OPTIMIZELY_CLIENT_ODP_SEGMENTSCACHE environment variable
+	if odpSegmentsCache := v.GetStringMap("client.odp.segmentsCache"); odpSegmentsCache != nil {
```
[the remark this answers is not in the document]
yasirfolio3: No specific reason, its just something that we did for `ups` aswell.

plugins/odpcache/services/in_memory_cache.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"time"
+
+	"github.com/optimizely/agent/plugins/odpcache"
+	"github.com/optimizely/go-sdk/pkg/odp/cache"
+)
+
+// InMemoryCache represents the in-memory implementation of Cache interface
+type InMemoryCache struct {
```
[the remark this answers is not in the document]
yasirfolio3: We are now using this as default odp cache with both `timeout` and `size` as arguments. Following is the default configuration:

```
odp:
      ## Disable odp
      disable: false
      ## Timeout in seconds after which event requests will timeout.
      eventsRequestTimeout: 10s
      ## Flush interval in seconds for odp events
      eventsFlushInterval: 1s
      ## Timeout in seconds after which segment requests will timeout.
      segmentsRequestTimeout: 10s
      ## This will override all other settings
      segmentsCache:
        default: "in-memory"
        services:
          in-memory: 
            size: 10000
            timeout: 600 # seconds
        #   redis: 
        #     host: "localhost:6379"
        #     password: ""
        #     database: 0
        #     timeout: 0 # seconds
```

config.yaml
```diff
@@ -157,6 +157,31 @@ client:
...
+      eventsRequestTimeout: 10s
+      ## Flush interval in seconds for odp events
+      eventsFlushInterval: 1s
+      ## Max number of items that can be stored in odp segments cache.
+      segmentsCacheSize: 10000
+      ## Timeout in seconds after which an item stored in odp segments cache will be deleted.
+      segmentsCacheTimeout: 600s
+      ## Timeout in seconds after which segment requests will timeout.
+      segmentsRequestTimeout: 10s
+      ## This will override all other settings
```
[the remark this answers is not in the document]
yasirfolio3: deleted this comment as `segmentsCacheTimeout ` and `segmentsCacheSize` no longer exist outside of the scope of `in-memory` cache.

config.yaml
```diff
@@ -157,6 +157,31 @@ client:
...
+      segmentsCache:
+        default: ""
```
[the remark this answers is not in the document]
yasirfolio3: as per the new implementation, empty default simply means that go-sdk will use its default LRU cache with default cache timeout and size

cmd/optimizely/testdata/default.yaml
```diff
@@ -65,6 +63,8 @@ client:
         redis: 
           host: "localhost:6379"
           password: ""
+          timeout: 5
```
jaeopt, line 66: 5s ?

config.yaml
```diff
@@ -164,23 +164,20 @@ client:
...
       segmentsCache:
-        default: ""
+        default: "in-memory"
         services:
-        #   in-memory: 
-        #     size: 0
-        #     timeout: 0 # seconds
+          in-memory: 
+            size: 10000
+            timeout: 600 # seconds
```
jaeopt, line 175: 600s ?

config.yaml
```diff
@@ -164,23 +164,20 @@ client:
...
         #   redis: 
         #     host: "localhost:6379"
         #     password: ""
         #     database: 0
+        #     timeout: 0 # seconds
```
jaeopt, line 180: 600s default?

config.yaml
```diff
@@ -164,23 +164,20 @@ client:
       eventsRequestTimeout: 10s
       ## Flush interval in seconds for odp events
       eventsFlushInterval: 1s
-      ## Max number of items that can be stored in odp segments cache.
-      segmentsCacheSize: 10000
-      ## Timeout in seconds after which an item stored in odp segments cache will be deleted.
-      segmentsCacheTimeout: 600s
       ## Timeout in seconds after which segment requests will timeout.
       segmentsRequestTimeout: 10s
       ## This will override all other settings
```
jaeopt, line 169: No overriding any more. This line should be removed.

jaeopt (changes requested):
All changes look good! A few small fixes suggested.

cmd/optimizely/testdata/default.yaml
yasirfolio3: `5s` was possible if the key was at the root and viper was unmarshalling it. In this case we are unmarshalling it at runtime using native `json` library which does not support converting `5s` to time duration.

config.yaml
yasirfolio3: same as above.

config.yaml
yasirfolio3: same as above.

plugins/odpcache/README.md
```diff
@@ -0,0 +1,65 @@
...
+```
+## configure optional ODP Cache
+odp:
+  segmentsCache:
+    default: "in-memory"
+    services:
+      in-memory: 
+        ## 0 means cache will be disabled
+        size: 0
+        ## timeout after which least recently used records will be deleted
```
jaeopt, line 16:
```suggestion
        ## timeout after which the cached item will become invalid.
```

plugins/odpcache/README.md
```diff
@@ -0,0 +1,65 @@
...
+Use a ODP Cache to cache user segments fetched from the ODP server.
+
+## Out of Box Cache Usage
+
+1. To use the in-memory `ODPCache`, update the `config.yaml` as shown below:
+```
+## configure optional ODP Cache
+odp:
+  segmentsCache:
+    default: "in-memory"
```
jaeopt, line 11: It can be better if default "default" is clarified. If no  `segmentsCache` is defined, we will use the default in-memory with default size and timeout?

plugins/utils/duration.go
```diff
@@ -0,0 +1,53 @@
...
+	var unmarshalledJSON interface{}
+
+	err := json.Unmarshal(b, &unmarshalledJSON)
+	if err != nil {
+		return err
+	}
+
+	switch value := unmarshalledJSON.(type) {
+	case float64:
+		duration.Duration = time.Duration(value)
```
jaeopt, line 42: Do we accept number without unit? What is the default unit for this case?

jaeopt (changes requested):
Looks great! A few minor fixes suggested.

plugins/utils/duration.go
yasirfolio3: yes we do, default unit is nanosecond by golang.

jaeopt (approved):
LGTM!

PR #366, opened 2023-03-06
FSSDK-8970: Adds support for odp options in config.

yasirfolio3:
## Summary
- Adds support for providing `disableOdp`, `eventsRequestTimeout`, `eventsFlushInterval`, `segmentsCacheSize`, `segmentsRequestTimeout` and `segmentsCacheTimeout` options through config.yaml

## Issues
- https://jira.sso.episerver.net/browse/FSSDK-8970

## Tests
- Added new unit tests.
- All previous tests should pass.


config.yaml
```diff
@@ -135,6 +135,12 @@ client:
     ## By default Agent assumes only alphanumeric characters as part of the SDK Key string.
     ## https://github.com/google/re2/wiki/Syntax
     sdkKeyRegex: "^\\w+(:\\w+)?$"
+    ## Disable odp
+    disableOdp: false
+    ## Max number of items that can be stored in odp segments cache.
+    segmentsCacheSize: 10000
+    ## Timeout in seconds after which an item stored in odp segments cache will be deleted.
+    segmentsCacheTimeout: 600s
```
jaeopt, line 143: What about grouping all these and future odp options under "odp". I think "odpCache" is also can be a part of this group. We can also clarify "odpCache" will override these values.

jaeopt (changes requested):
I have a suggestion for grouping all odp configurations. Let me know if it makes sense.
Other than that, all looks good to me.

config.yaml
Mat001: What about:
-  odp_segment_request_timeout
- odp_event_request_timeout
- odp_event_flush_interval

We use these in Python sdk ODP sdk settings, but don't see them exposed in go-sdk. Are we including them in agent config?

Mat001 (changes requested):
Made a comment about config options.

config.yaml
yasirfolio3: Have only added those params which are directly available through the main factory constructor. Should we be adding these internal properties as well? @jaeopt

yasirfolio3: @Mat001 request timeout's and flush interval can also be updated in go-sdk but for that we need to supply instances of both the event manager and segment manager to the odp manager.

jaeopt: @Mat001 a good call! I think it's valuable to provide those options (segment_request_timeout, event_request_timeout, and probably event_flush_interval as well) to agent customers. @yasirfolio3

config.yaml
```diff
@@ -164,6 +158,13 @@ client:
...
         #     Auth-Token: "12345"
     odp:
+      ## Disable odp
+      disable: false
+      ## Max number of items that can be stored in odp segments cache.
+      segmentsCacheSize: 10000
+      ## Timeout in seconds after which an item stored in odp segments cache will be deleted.
+      segmentsCacheTimeout: 600s
+      ## This will override all other settings
       cache:
```
jaeopt, line 168: Looks good! What about "segmentsCache" instead of "cache" to be consistent with "segmentsCacheSize" and "segmentsCacheTimeout"?

jaeopt (changes requested):
Looks good! A name change suggested. Can we also add timeouts and flush-interval to odp options as @Mat001 suggested?

config.yaml
yasirfolio3: sure, will update this in the parent PR for odp-cache, this is branched off of it.

jaeopt (approved):
LGTM

Mat001 (approved):
Looks good with the additional config options.

PR #368, opened 2023-03-10
[FSSDK-8958] Change Full Stack to Feature Experimentation

griffincox-optimizely:
Change Full Stack to Feature Experimentation


griffincox-optimizely:
Consider removing entire Readme Sync directory per conversation in FX Rename: SDK Swarm Day

PR #383, opened 2023-03-23
[FSSDK-8839]: Added acceptance tests for send-odp-event.

yasirfolio3:
## Summary
- Adds acceptance tests or send-odp-event api.
- Fixes a crash caused by nil logger being passed with odp components.

## Issues
- [FSSDK-8839](https://jira.sso.episerver.net/browse/FSSDK-8839)

## Tests
- All tests should pass

### DIFF
--- a/tests/acceptance/test_acceptance/test_send_odp_event.py
+++ b/tests/acceptance/test_acceptance/test_send_odp_event.py
@@ -1,0 +1,210 @@
+import json
+
+import pytest
+import requests
+
+from tests.acceptance.helpers import ENDPOINT_SEND_ODP_EVENT
+from tests.acceptance.helpers import create_and_validate_request_and_response
+
+expected_odp_not_integrated = {
+    'error': 'ODP is not integrated'
+}
+
+expected_missing_identifiers = {
+    'error': 'missing or empty "identifiers" in request payload'
+}
+
+expected_invalid_data = {
+    'error': 'ODP data is not valid'
+}
+
+expected_missing_action = {
+    'error': 'missing "action" in request payload'
+}
+
+expected_request_error = {
+    'error': 'error parsing request body'
+}
+
+expected_successful_odp_event = {
+    "success": True
+}
+
+
+@pytest.mark.parametrize(
+    "expected_response, expected_status_code", [
+        (expected_odp_not_integrated, 500)
+    ])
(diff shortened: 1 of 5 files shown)


tests/acceptance/test_acceptance/test_send_odp_event.py
```diff
@@ -0,0 +1,210 @@
...
+
+        assert json.loads(json.dumps(expected_response)) == resp.json()
+        assert resp.status_code == expected_status_code
+        assert resp.text == expected_error
+        resp.raise_for_status()
+
+
+@pytest.mark.parametrize(
+    "expected_response, expected_status_code, expected_error, bypass_validation_request, bypass_validation_response", [
+        (expected_invalid_data, 500,
```
jaeopt, line 147: Do we return 500 (not 400) for this user error?

jaeopt (changes requested):
Looks good! A clarification about error status code.

tests/acceptance/test_acceptance/test_send_odp_event.py
yasirfolio3: yes, in case an error is returned from the sdk side, we return 500.

yasirfolio3: https://github.com/optimizely/agent/blob/ce3b3a708a84d844dc3548b866f973317c3276c8/pkg/handlers/send_odp_event.go#L47

yasirfolio3: [same for ](https://github.com/optimizely/agent/blob/ce3b3a708a84d844dc3548b866f973317c3276c8/pkg/handlers/track.go#L65)

yasirfolio3: > It makes sense. But it's a kind of user request errors. Wondering how we handle other errors when go-sdk (not agent) detects any error in api parameters (like invalid attributes for `decide` api). Do we return 500?

if we validate parameters before executing request, we send 400, but after request execution we return 500. In this case we are not validating separately if `data` only had types that we support, we are counting on go-sdk to do that.

jaeopt (approved):
LGTM

Mat001 (approved):
LGTM

PR #384, opened 2023-04-03
[FSSDK-9026](WIP): Adds tests for redis cache.

yasirfolio3:
## Summary
- Adds acceptance tests for redis odp cache.

## Issues
- [FSSDK-9026](https://jira.sso.episerver.net/browse/FSSDK-9026)

## Tests
- All tests should pass

### DIFF
--- a/tests/acceptance/test_acceptance/test_odp_redis.py
+++ b/tests/acceptance/test_acceptance/test_odp_redis.py
@@ -1,0 +1,148 @@
+import json
+
+import pytest
+import requests
+import redis
+import time
+
+from tests.acceptance.helpers import ENDPOINT_DECIDE
+from tests.acceptance.helpers import create_and_validate_request_and_response
+from tests.acceptance.helpers import sort_response
+
+expected_redis_save = {
+    "variationKey": "variation_b",
+    "enabled": True,
+    "ruleKey": "ab_experiment",
+    "flagKey": "flag1",
+    "userContext": {
+        "userId": "fs-id-1",
+        "attributes": {}
+    },
+    "reasons": ["Audiences for experiment ab_experiment collectively evaluated to true."]
+}
+
+expected_redis_lookup = {
+    "variationKey": "variation_b",
+    "enabled": True,
+    "ruleKey": "ab_experiment",
+    "flagKey": "flag1",
+    "userContext": {
+        "userId": "fs-id-2",
+        "attributes": {}
+    },
+    "reasons": ["Audiences for experiment ab_experiment collectively evaluated to true."]
+}
+
+expected_redis_reset = {
+    "variationKey": "variation_a",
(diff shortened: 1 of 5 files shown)


tests/acceptance/test_acceptance/test_odp_redis.py
```diff
@@ -0,0 +1,150 @@
...
+    "reasons": ["Audiences for experiment ab_experiment collectively evaluated to true."]
+}
+
+def test_redis_save(session_override_sdk_key_odp):
+    """
+    Test that first fetch call saves segments in redis successfuly.
+    :param session_override_sdk_key_odp: sdk key to override the session using odp key
+    """
+    
+    expected_segments = ["atsbugbashsegmentdob", "atsbugbashsegmentgender"]
```
Mat001, line 54: Yasir, remind me which project are we using for this test? Did we create a new project in QA account? And you set sdk key from there right? 
This test is not tied to my personal project?

We also need to make sure that segments "atsbugbashsegmentdob", "atsbugbashsegmentgender" remain in ODP and nobody removes them for these tests to work. That's probably on me.

Mat001: I also see we are using "fs-id-1". I have noticed some inconsistencies in ODP platform with how segments were returned with fs-id-xy users. Maybe some other folks used this id as well. This issue is going to be discussed with ODP team as part of the ATS bug bash. PS: not sure if using "matjaz-user-1" would be safer, perhaps, can discuss.

Let's hope that returned segments remain consistent.

tests/acceptance/test_acceptance/test_odp_redis.py
```diff
@@ -0,0 +1,150 @@
...
+    uId = "fs_user_id-$-fs-id-2"
+    r = redis.Redis(host='localhost', port=6379, db=0)
+    r.set(uId, json.dumps(expected_segments))
+
+    payload = {
+        "userId": "fs-id-2",
```
Mat001, line 94: Same for here.

Mat001 (approved):
Looks good. Minor comment.

tests/acceptance/test_acceptance/test_odp_redis.py
yasirfolio3: > Yasir, remind me which project are we using for this test? Did we create a new project in QA account? And you set sdk key from there right? This test is not tied to my personal project?
> 
> We also need to make sure that segments "atsbugbashsegmentdob", "atsbugbashsegmentgender" remain in ODP and nobody removes them for these tests to work. That's probably on me.

we are using a separate project from the QA account, the same that we created for ATS tests. Yes we do have to make sure that the segments are not deleted.

Mat001 (approved):
Looks good

PR #390, opened 2023-05-10
fix(github-actions): fix sourceclear github actions failure

pulak-opti:
## Summary
- There is an unreachable forced [import](https://github.com/optimizely/agent/blob/master/pkg/routers/api.go#L31) which causes failure in github action in sourceclear part. For this sourceclear failed to list dependencies using `go list -deps -json all` command.

- The local statik package is ignored in the .gitignore & only available after build. Without building the project `go mod tidy` command couldn't find the local statik package.


yasirfolio3 (approved):
LGTM

PR #392, opened 2023-05-16
[FSSDK-8978] chore: update openapi schema to 3.1.0

pulak-opti:
## Summary
- Update Open API schema to 3.1.0

## Ticket
- [FSSDK-8978](https://jira.sso.episerver.net/browse/FSSDK-8978)


tests/acceptance/helpers.py
```diff
@@ -163,22 +136,39 @@ def create_and_validate_request_and_response(endpoint, method, session, bypass_v
...
     )
 
-    if not bypass_validation_request:
-        request_result.raise_for_errors()
 
-    base_url = os.getenv('host')
+  
+    try:
+        create_and_validate_request(request)
+    except Exception as e:
```
Mat001, line 155: Could we try to narrow down which exceptions might occur and use those? instead of a generic Exception.

tests/acceptance/helpers.py
```diff
@@ -163,22 +136,39 @@ def create_and_validate_request_and_response(endpoint, method, session, bypass_v
...
-    response_result = create_and_validate_response(request, response)
-
-    if not bypass_validation_response:
-        response_result.raise_for_errors()
 
+    try: 
+        create_and_validate_response(request, response)
+    except MissingData:
+        return response
+    except Exception as e:
```
Mat001, line 168: Same here. Would be easier to debug as error messages would be more specific. An example here: https://github.com/optimizely/python-sdk/blob/master/optimizely/odp/odp_segment_api_manager.py

But if that's too complicated we could just keep a generic Exception, this code is not super critical or super customer facing.

Mat001 (changes requested):
Lookd great. Just a request if you can see if Exception can be more specific. If too complicated, then we can keep Exceptipn.

tests/acceptance/helpers.py
pulak-opti: Thanks @Mat001 for the suggestions
updated with specific exceptions now

tests/acceptance/helpers.py
pulak-opti: same here as well, updated with specific exceptions

Mat001 (approved):
Looks good!

PR #395, opened 2023-07-25
[FSSDK-9512] chore: add a warning log when polling interval is below threshold

pulak-opti:


pulak-opti:
Not needed anymore as the warning log is already implemented in go-sdk https://github.com/optimizely/go-sdk/pull/378

PR #399, opened 2023-09-12
[FSSDK-9632] feat: Implement notification consistency across agent nodes

pulak-opti:
## Summary
- Implement notification consistency across agent nodes

## Tickets
- [FSSDK-9632](https://jira.sso.episerver.net/browse/FSSDK-9632)


pkg/syncer/syncer.go
```diff
@@ -0,0 +1,134 @@
+/****************************************************************************
+ * Copyright 2019-2020,2023, Optimizely, Inc. and contributors              *
```
mikechu-optimizely, line 2: Should this have only 2023?

pkg/handlers/notification_test.go
```diff
@@ -170,11 +245,19 @@ func (suite *NotificationTestSuite) TestActivateExperimentRaw() {
...
 	nc := registry.GetNotificationCenter("")
+	decisionEvent := map[string]string{"key": "value"}
+
+	notifications := make([]syncer.Event, 0)
+	notifications = append(notifications, syncer.Event{Type: notification.Decision, Message: decisionEvent})
+
 	go func() {
 		time.Sleep(1 * time.Second)
-		nc.Send(notification.Decision, map[string]string{"key": "value"})
+		nc.Send(notification.Decision, decisionEvent)
```
mikechu-optimizely, line 255: Much nicer 👍

mikechu-optimizely (approved):
Transparently: I do not completely understand the implementation details, but I trust your tests cover the functional additions.  LGTM

config.yaml
```diff
@@ -205,3 +205,17 @@ runtime:
...
+synchronization:
+    notification:
+        enable: false
+        default: "redis"
+        pubsub:
+            redis:
+                host: "redis.demo.svc:6379"
+                password: ""
+                database: 0
+                channel: "optimizely-notifications"
```
jaeopt, line 221: What about this, so that we can use `pubsub` for any synchronization, not only for notification?

```suggestion
synchronization:
    pubsub:
        redis:
            host: "redis.demo.svc:6379"
            password: ""
            database: 0
            channel: "optimizely-notifications"
    notification:
        enable: false
        default: "redis"
    datafile:
        enable: false
        default: "redis"    
```

pkg/handlers/notification.go
```diff
@@ -51,112 +64,202 @@ func getFilter(filters []string) map[string]notification.Type {
...
+		go func() {
+			for {
+				select {
+				case <-ctx.Done():
+					client.Close()
+					pubsub.Close()
+					logger.Debug().Msg("context canceled, redis notification receiver is closed")
+					return
+				default:
+					msg, err := pubsub.ReceiveMessage(ctx)
```
jaeopt, line 247: It looks like we're getting notifications for all sdkKeys. If so, we need to filter out those notifcations for the requested sdkKey only. Can we consider a channel per sdkKey?

jaeopt (changes requested):
Looks great! 
A couple of changes suggested. I will review again after discussion.

pkg/handlers/notification.go
```diff
@@ -51,112 +64,202 @@ func getFilter(filters []string) map[string]notification.Type {
...
 	nc := registry.GetNotificationCenter(sdkKey)
 
-	// Parse the form.
-	_ = r.ParseForm()
-
-	filters := r.Form["filter"]
-
 	// Parse out the any filters that were added
-	notificationsToAdd := getFilter(filters)
+	notificationsToAdd := types
```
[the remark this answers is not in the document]
pulak-opti: Here we had to change the logic a bit, now we are looking for any kind of notifications and do the filtering before writing to the response body.

config.yaml
pulak-opti: looks good, we might need a different channel for each service i.e. notification, datafile etc.

pkg/handlers/notification.go
pulak-opti: Yes! we can maintain a pub-sub channel for each sdk key.

pulak-opti: added sdk key based filtering

pkg/routers/api.go
```diff
@@ -63,35 +63,37 @@ func forbiddenHandler(message string) http.HandlerFunc {
...
 		overrideHandler = forbiddenHandler("Overrides not enabled")
 	}
 
-	nStreamHandler := handlers.NotificationEventSteamHandler
-	if !conf.EnableNotifications {
-		nStreamHandler = forbiddenHandler("Notification stream not enabled")
+	nStreamHandler := forbiddenHandler("Notification stream not enabled")
+	if conf.API.EnableNotifications {
+		nStreamHandler = handlers.NotificationEventStreamHandler(handlers.DefaultNotificationReceiver)
+		if conf.Synchronization.Notification.Enable {
```
jaeopt, line 87: Do we need to enable both to turn on notification stream? What about sync.notification.enable overrides?

jaeopt (changes requested):
Looks good with the sdkKey-based channels!
I want to consider a sync-er shared for all synchronizations supporting different types of events. We can chat about the options.

config.yaml
```diff
@@ -249,12 +249,12 @@ runtime:
 ## if notification synchronization is enabled, then the active notification event-stream API
 ## will get the notifications from multiple replicas
 synchronization:
+    pubsub:
+        redis:
+            host: "redis.demo.svc:6379"
+            password: ""
+            database: 0
+            channel: "optimizely-notifications"
```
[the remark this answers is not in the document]
pulak-opti: what about "optimizely-sync"?

config.yaml
```diff
@@ -249,12 +249,12 @@ runtime:
...
     notification:
         enable: false
         default: "redis"
```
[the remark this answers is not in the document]
pulak-opti: Is there any chance of having multiple options like pubsub in future?

pkg/routers/api.go
pulak-opti: Yes, top level API.EnableNotifications means Notification API is enabled/disabled. then if it is enabled, then there are two choices. By default, synchronization is disabled.

If the config is changed or overridden, the server needs to be restarted to get the effect.

config.yaml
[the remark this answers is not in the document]
jaeopt: For name conflicts, I doubt. Just for logical flow. Up to you.

pkg/syncer/syncer.go
```diff
@@ -0,0 +1,164 @@
...
+var (
+	ncCache   = make(map[string]*RedisNotificationSyncer)
+	mutexLock = &sync.Mutex{}
+)
+
+// Event holds the notification event with it's type
+type Event struct {
+	Type    notification.Type `json:"type"`
+	Message interface{}       `json:"message"`
+}
```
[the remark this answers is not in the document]
pulak-opti: Yes, now we can but we can't change the Send() method parameter as we need to implement the notification [Center](https://github.com/optimizely/go-sdk/blob/master/pkg/notification/center.go#L29) interface. So we might have to maintain different struct for different sync use case.

pkg/syncer/syncer.go
```diff
@@ -59,8 +59,8 @@ type RedisNotificationSyncer struct {
 	sdkKey   string
 }
 
-// NewRedisNotificationSyncer returns an instance of RedisNotificationSyncer
-func NewRedisNotificationSyncer(logger *zerolog.Logger, conf config.SyncConfig, sdkKey string) (*RedisNotificationSyncer, error) {
+// NewRedisSyncer returns an instance of RedisNotificationSyncer
+func NewRedisSyncer(logger *zerolog.Logger, conf config.SyncConfig, sdkKey string) (*RedisSyncer, error) {
```
jaeopt, line 63: Looks good. I had something like this in mind. We can share the same syncer.
I approve it as is. We can discuss more after datafile-syncer is added. Then we may have a better idea :)

```
NewNotificationHandler(syncer: RedisSyncer, config, sdkKey) {
    ...
    send(event) {
        syncer.send(event.message)
    }
}
```

jaeopt (approved):
LGTM - I want to discuss more about our options after we integrate datafile syncer.

PR #400, opened 2023-10-05
[FSSDK-9705] feat: add OpenTelemetry tracing

pulak-opti:
## Summary
- Add OpenTelemetry tracing

## Ticket
- [FSSDK-9705](https://jira.sso.episerver.net/browse/FSSDK-9705)


mikechu-optimizely (approved):
Mostly just questions 👇 but otherwise LGTM...though probably should also get @jaeopt 's review.

config.yaml
```diff
@@ -26,6 +26,22 @@ log:
...
+        serviceName: "optimizely-agent"
+        env: "dev"
+        services:
+            stdout:
+                filename: "trace.out"
+            remote:
+                endpoint: "localhost:4317"
+                ## supported protocals are "http" and "grpc"
+                protocal: "grpc"
+                sampleRate: 1.0
```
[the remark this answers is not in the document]
mikechu-optimizely: Well this was immediately outdated. 😁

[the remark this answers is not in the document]
pulak-opti: I just finished it and am checking your review 😁

go.mod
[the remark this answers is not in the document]
pulak-opti: A minor version update of Agent will be a viable choice. Also we need to release a new version of helm chart with the updated config.

cmd/optimizely/main.go
```diff
@@ -140,6 +239,22 @@ func main() {
 	conf := loadConfig(v)
 	initLogging(conf.Log)
 
+	if conf.Tracing.Enabled {
```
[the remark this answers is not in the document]
pulak-opti: We need to provided necessary info via config.yaml. and in start-up all is set up before running the server.

pkg/routers/api.go
```diff
@@ -144,19 +144,19 @@ func WithAPIRouter(opt *APIOptions, r chi.Router) {
...
-		r.With(decideTimer, opt.oAuthMiddleware, contentTypeMiddleware).Post("/decide", opt.decideHandler)
-		r.With(trackTimer, opt.oAuthMiddleware, contentTypeMiddleware).Post("/track", opt.trackHandler)
-		r.With(overrideTimer, opt.oAuthMiddleware, contentTypeMiddleware).Post("/override", opt.overrideHandler)
-		r.With(lookupTimer, opt.oAuthMiddleware, contentTypeMiddleware).Post("/lookup", opt.lookupHandler)
-		r.With(saveTimer, opt.oAuthMiddleware, contentTypeMiddleware).Post("/save", opt.saveHandler)
-		r.With(sendOdpEventTimer, opt.oAuthMiddleware, contentTypeMiddleware).Post("/send-odp-event", opt.sendOdpEventHandler)
-		r.With(opt.oAuthMiddleware).Get("/notifications/event-stream", opt.nStreamHandler)
+		r.With(getConfigTimer, opt.oAuthMiddleware, middleware.AddTracing("configHandler", "OptimizelyConfig")).Get("/config", opt.configHandler)
+		r.With(getDatafileTimer, opt.oAuthMiddleware, middleware.AddTracing("datafileHandler", "OptimizelyDatafile")).Get("/datafile", opt.datafileHandler)
+		r.With(activateTimer, opt.oAuthMiddleware, contentTypeMiddleware, middleware.AddTracing("activateHandler", "Activate")).Post("/activate", opt.activateHandler)
```
jaeopt, line 149: Wondering why we do tracing here and also in ActivateHandler redundantly?

pkg/middleware/trace.go
```diff
@@ -0,0 +1,36 @@
...
+	"net/http"
+
+	"go.opentelemetry.io/otel"
+)
+
+func AddTracing(tracerName, spanName string) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		fn := func(w http.ResponseWriter, r *http.Request) {
+			ctx, span := otel.Tracer(tracerName).Start(r.Context(), spanName)
+			defer span.End()
```
jaeopt, line 30: This middleware solution looks good. Wondering any way we can add attributes consistently for all APIs.

pkg/middleware/trace.go
```diff
@@ -0,0 +1,36 @@
...
+
+			next.ServeHTTP(w, r.WithContext(ctx))
```
jaeopt, line 32: I do not see how to extract or add trace id for distributed tracing. Can we add it here?

jaeopt (changes requested):
Looks great! A few suggestions and clarifications.

config.yaml
```diff
@@ -26,6 +26,41 @@ log:
     ## to set whether or not the SDK key is included in the logging output.
     includeSdkKey: true
 
+##
+## tracing: tracing configuration
+##
+tracing:
```
jaeopt, line 32: What about "opentelemetry"? I do not strong opinion on this, but this tracing config defintion is for OTEL only and the name will make it clear.
Or we can also consider "tracing > opentelemetry" for extensibility.

config.yaml
```diff
@@ -26,6 +26,41 @@ log:
...
+        ## tracing service configuration
+        services:
+            ## stdout exporter configuration
+            stdout:
+                ## for stdout tracing data is saved in the specified file
+                filename: "trace.out"
+            ## remote exporter configuration
+            remote:
+                ## remote collector endpoint
+                endpoint: "localhost:4317"
```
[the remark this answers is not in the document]
pulak-opti: Yes, this can be connected with any trace collector. I have tested with Jaeger locally. I can make sure the same config will work for datadog as well.

As this is working as push-model (agent will push trace data). Here remote means a remote trace collector which can be any vendor that's why skipping any vendor name.

pkg/routers/api.go
pulak-opti: Actually from Activate Handler, we are calling some more important method. We are covering those too. It'll provide trace data like a chain.

config.yaml
pulak-opti: Yeah, love the idea. But OpenTelemetry actually provides all tracing, metrics, logging. We can also think about "opentelemetry" > "tracing". 

About extensibility, I think the currect model is extensible in terms of the trace data that can be exported anywhere. But we want to collect trace data in many ways then extensibility comes into the picture. But collecting the same trace data in many ways might not be that important.

would love discuss more on this

pkg/optimizely/client.go
```diff
@@ -83,7 +86,11 @@ func (c *OptlyClient) UpdateConfig() {
 }
 
 // TrackEvent checks for the existence of the event before calling the OptimizelyClient Track method
-func (c *OptlyClient) TrackEvent(eventKey string, uc entities.UserContext, eventTags map[string]interface{}) (*Track, error) {
+func (c *OptlyClient) TrackEvent(ctx context.Context, eventKey string, uc entities.UserContext, eventTags map[string]interface{}) (*Track, error) {
+	_, span := otel.Tracer("trackHandler").Start(ctx, "TrackEvent")
+	defer span.End()
+	span.SetAttributes(attribute.String("trackEventKey", eventKey))
```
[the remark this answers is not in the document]
pulak-opti: ah, seems I missed it. Yeah, though in some API like /save, /config, seems there is no important attribute to collect. But I'll look for more and add those.

pkg/middleware/trace.go
pulak-opti: collecting attributes from middleware would be difficult, I think. Also in some API, it seems there is hardly any attributes (extra info) to collect.

pkg/middleware/trace.go
pulak-opti: It seems, every trace (in this case every API call) has (should be) a different trace id and it is the ideal otherwise each trace can't be identified separately.

pkg/middleware/trace.go
pulak-opti: btw added some common http attributed in the middleware

config.yaml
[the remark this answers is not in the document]
jaeopt: Authorization token for vendors?

pkg/routers/api.go
jaeopt: I now see the point.  Not sure about the value of adding tracing for the sub levels since they won't care or may be able to differenticate from attributes?

pkg/middleware/trace.go
jaeopt: My question is that if upstream services add trace-id to agent requests, we need to extract it and add to our traces.

pkg/middleware/trace.go
pulak-opti: Most of the APIs don't have it. Also it seems those attributes are just considered as extra information in tracing sounds like less important things (not in our case btw :) ). Maybe we can leave it for now.

pkg/middleware/trace.go
pulak-opti: The OpenTelemetry itself generates a random number and sets it as trace id actually.

config.yaml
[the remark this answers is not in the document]
pulak-opti: Yeah, I also looked for it inside OpenTelemetry. But surprising don't see any such functionality to pass such token so far. I think, agent (any tracing enabled server) & trace collector (example: Jaeger) will run inside private network and might not need any such token or security. But collector UI i.e. to see/explore those data can need token. that will depend on that vendor settings.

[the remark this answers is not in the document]
pulak-opti: I have found these available options [here](https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/otlp/otlptrace/otlptracehttp/options.go#L63-L116)

pkg/middleware/trace.go
```diff
@@ -21,15 +21,43 @@ import (
...
 			ctx, span := otel.Tracer(tracerName).Start(r.Context(), spanName)
 			defer span.End()
 
-			next.ServeHTTP(w, r.WithContext(ctx))
+			span.SetAttributes(
+				semconv.HTTPMethodKey.String(r.Method),
+				semconv.HTTPRouteKey.String(r.URL.Path),
+				semconv.HTTPURLKey.String(r.URL.String()),
+				semconv.HTTPHostKey.String(r.Host),
+				semconv.HTTPSchemeKey.String(r.URL.Scheme),
```
jaeopt, line 48: What about sdkKey in request header?

pulak-opti: added

jaeopt (changes requested):
All looks good! A nit suggested.
I still see the "tracing" instead of "opentelemetry: tracing" in the config.yml.
Wondering about the plan.

pkg/middleware/trace.go
```diff
@@ -18,12 +18,65 @@
...
+	"sync"
 
+	"github.com/rs/zerolog/log"
 	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
 	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+	"go.opentelemetry.io/otel/trace"
 )
 
+type traceIDGenerator struct {
```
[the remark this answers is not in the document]
pulak-opti: This struct is intentionally left as Private. this struct has implemented an interface and has a public New...() func to work with. it.

pulak-opti:
> All looks good! A nit suggested. I still see the "tracing" instead of "opentelemetry: tracing" in the config.yml. Wondering about the plan.

ah sorry, config changes are fall behind for distributed tracing. btw I am still thinking what the config should be. 
now tracing > opentelemetry makes more sense if we have other tracing implementation in future.

jaeopt (approved):
LGTM

PR #401, opened 2023-10-26
fix: fix trace middleware http writer

pulak-opti:
## Summary
- use http response writer from chi router to avoid error case like notification api expects this writer to implement flusher interface. Using http response writer from chi router eliminate any such errors.

- fixed one continuously failing acceptance test


jaeopt (approved):
LGTM

PR #402, opened 2023-11-01
fix: fix context propagation in distributed tracing

pulak-opti:
## Summary
- fix context propagation in distributed tracing
- `traceparent` header is used to get tracing context


jaeopt (approved):
LGTM

PR #406, opened 2023-11-17
adds cache to ups lookup call

204Constie:
## Summary
the service has very poor performance when one request is done for multiple experiments at once
the PR adds caching on lookup method to prevent async request loop when not necessary

## Issues
- improves performance in case a single request is done for multiple experiments at once


last-ent (approved):

pulak-opti (changes requested):
Shared some thoughts.

PR #416, opened 2024-02-22
[FSSDK-9972]  fix dependabot moderate alerts

pulak-opti:
## Summary
- updated dependencies to fix [dependabot moderate alerts](https://github.com/optimizely/agent/security/dependabot)
- github.com/lestrrat-go/jwx library introduced breaking changes to fix the vulnerability, so updated accordingly

## Ticket
- [FSSDK-9972](https://jira.sso.episerver.net/browse/FSSDK-9972)


pkg/middleware/auth.go
```diff
@@ -130,15 +131,15 @@ func (c *JWTVerifierURL) updateKeySet() error {
 	c.jwksLock.Lock()
 	defer c.jwksLock.Unlock()
 
-	set, err := jwk.Fetch(c.jwksURL)
+	set, err := jwk.Fetch(context.TODO(), c.jwksURL)
```
mikechu-optimizely, line 134: I'm totally not knowledgeable, but `.TODO` caught my eye. This is correct right?

mikechu-optimizely (approved):
Small thing that caught my eye 👇 but LGTM

raju-opti (approved):
LGTM

pkg/middleware/auth.go
pulak-opti: Yes, for now, we have to pass a context there. TODO() is for in future we should pass the caller context here if possible.

PR #420, opened 2024-07-18
[FSSDK-10397] remove redundant info log & replace info log with debug log inside loop

pulak-opti:
## Summary
- removed redundant info logs
- replace info log with debug log inside loop

## Issues
- [FSSDK-10397](https://jira.sso.episerver.net/browse/FSSDK-10397)


mikechu-optimizely (approved):
LGTM

PR #422, opened 2024-08-20
[FSSDK-10553] add IsEveryoneElseVariation marker in decide API

pulak-opti:
## Summary
- added IsEveryoneElseVariation marker in decide API

## Ticket
- [FSSDK-10553](https://jira.sso.episerver.net/browse/FSSDK-10553)

### DIFF
--- a/pkg/handlers/decide.go
+++ b/pkg/handlers/decide.go
@@ -123,7 +123,7 @@ func Decide(w http.ResponseWriter, r *http.Request) {
 		key := keys[0]
 		logger.Debug().Str("featureKey", key).Msg("fetching feature decision")
 		d := optimizelyUserContext.Decide(key, decideOptions)
-		decideOut := DecideOut{d, d.Variables.ToMap(), IsEveryoneElseVariation(featureMap[d.FlagKey].DeliveryRules, d.RuleKey)}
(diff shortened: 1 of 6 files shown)


mikechu-optimizely (commented):
I read the Jira on this, but 2 initial thoughts:
1. Should this be solved in the SDK or via project configuration, as you mentioned as a variation?
2. If this is something that an SDK should handle, then perhaps it needs to be put in all the base language SDKs. 

There's a high probability you've already run through all these thoughts, and I'm late to understanding.

I do not see any problems with the coded solution, but will refrain from providing Approval since I'm not fully knowledgeable on the problem.

cc @jaeopt

pulak-opti:
Hey @mikechu-optimizely 
Thanks for your thoughts. I have tried to implement this in the SDK first. You might have noticed the linked closed PR in SDK. But then I have changed my mind for the followings:
- User has requested the feature for the Agent
- Implementing this in SDK will force us to add this in other primary language SDKs to keep things in sync.
- Users can implement the same thing on their server side in the similar way as we have done it in Agent if they need it.

jaeopt (approved):
LGTM

jaeopt:
@mikechu-optimizely @pulak-opti 
I also see some value for adding this to all SDKs, hiding our naming rule for everyone-else.
The current implementation in agent looks reasonable for now and we can review it again for adding into the SDK level.

PR #426, opened 2024-10-03
[FSSDK-10734] update regex to allow `=` character

pulak-opti:
## Summary
- updated config.yaml
- updated regex to allow `=` character

## Issues
- [FSSDK-10734](https://jira.sso.episerver.net/browse/FSSDK-10734)


mikechu-optimizely (approved):
LGTM. 

I like your use of character classes.  I would have used grouping with the OR `^(\\w|=)+(:(\\w|=)+)?$` but I'm a regex newbie :-)

Mat001 (approved):
lgtm

